PR #50, opened 2019-04-11
Only do releases from a branch staring with 'release'.

osery:
The plan is to have `master` (as opposed to `dev`) as the main branch but only allow SNAPSHOTS built there. The actual releases would get built in branch `release` and later on possibly in dedicated versioned release branches such as `release-1.0.0`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -105,8 +105,9 @@ pipeline {
 }
 
 /**
- * @return Version specified in gradle.properties. Fails if master contains a release version (to prevent pushing
- * release version accidentally).
+ * Returns version specified in gradle.properties.
+ *
+ * Fails if master contains a release version (to prevent pushing release version accidentally).
  */
 def getVersion() {
     def version = sh(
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -123,12 +115,29 @@ def getVersion() {
     if (!version) {
         error "Version must be set"
     }
+    if (isMasterBranch() && !isSnapshotVersion(version)) {
```
michal-raska: Should there be a check for snapshot preset in release branch, when we are checking for absence of snapshot in master?

osery: Good question. I was thinking it may make sense to have a release (e.g. after code-freeze) for testing but not release from it yet. Wdyt?

Jenkinsfile
```diff
@@ -105,16 +105,8 @@ pipeline {
...
- * @return true, if given version does not contain SNAPSHOT.
- */
-def isRelease(version) {
-    echo version
-    return !version.contains('SNAPSHOT')
-}
-
-/**
- * @return version specified in gradle.properties
+ * @return Version specified in gradle.properties. Fails if master contains a release version (to prevent pushing
```
zoido: nit: If this is javadoc, then Verbose main fragment is preferred before naked `@return` statement.

Jenkinsfile
michal-raska: Makes sense. Then maybe I would add a checkbox `Release`. If it's marked, it has to be a non-snapshot. Otherwise it must be a snapshot. I mean it's not a deal breaker I was just curious since your are checking it "one way", it might be worth checking the "other way" as well.

osery: That would mean one more step to actually release though. Let's try as is and we can add this additional step later when needed? Tbh, I would like to trigger the release from Jenkins anyway, just don't know how to update the versions etc.

Jenkinsfile
osery: Thanks. Done.

PR #55, opened 2019-04-16
Create aws s3 bucket outside of template

jahodfra:
This change is required, because terraform cannot adopt existing buckets outside of `us-east-1`.
h2oai/h2oai#8249

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-version = 0.1.2-SNAPSHOT
+version = 0.1.1-SNAPSHOT
 
 # Versions of dependencies. Try to keep these at the same version across the deployment templates to facilitate
 # issue resolution.
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-version = 0.1.1
+version = 0.1.2-SNAPSHOT
 
 # Versions of dependencies. Try to keep these at the same version across the deployment templates to facilitate
 # issue resolution.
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version = 0.1.1-SNAPSHOT
+version = 0.1.1
```
osery: Please don't bump version in the feature PR. Let's have releases in separate PRs for clarity of git history.
Also, let's keep master head always on `-SNAPSHOT`

osery:
What is the plan on this one? Shall we add this to the deployment dialog or use our fixed name bucket and boto3 to ensure the bucket exists first?

Did you consider `terraform import` as an alternative solution?

osery:
Something like: `https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#import`

jahodfra:
> What is the plan on this one? Shall we add this to the deployment dialog or use our fixed name bucket and boto3 to ensure the bucket exists first?
> 
> Did you consider `terraform import` as an alternative solution?

I already implemented solution in boto3 and I need this template before preparing PR into h2oai.
I'm not sure how the `terraform import` would behave if the bucket doesn't exist. Boto3 solution handles that.

osery:
I see. Please see the comment about the version.

gradle.properties
jahodfra: Done

PR #59, opened 2019-05-10
update mojo runtime in support-1.6 branch

thirdwing:
I am not sure this is the correct way to update mojo runtime and make a new release for DAI.

Please let me know if there are other things I need to do.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,7 +4,7 @@ version = 0.0.8-SNAPSHOT
 # issue resolution.
 
 # Internal dependencies:
-mojoRuntimeVersion = 0.13.18
+mojoRuntimeVersion = 0.13.20
 
 # External dependencies:
 awsLambdaCoreVersion = 1.2.0


osery:
@thirdwing ideally, we make the change separate from the release for nicer git history and archeology.
That would be this PR without the version change.

Next, we need to bump the version in the `support-1.6` branch to `SNAPSHOT-0.0.10` and create a branch `release-0.0.9` with a commit setting version to `0.0.9` (no snapshot here).

Normally, I tend to do this by two commits in the support-1.6 branch, one setting version to `0.0.9` and another setting it to `SNAPSHOT-0.0.10`. And then creating the `release-0.0.9` branch pointing to the first of the two commits. That will trigger the release process. Last, I also do a GitHub release (manually in the GitHub UI).

thirdwing:
OK. I will do that.

gradle.properties
```diff
@@ -4,7 +4,7 @@ version = 0.0.8-SNAPSHOT
 # issue resolution.
 
 # Internal dependencies:
-mojoRuntimeVersion = 0.13.18
+mojoRuntimeVersion = 0.13.20
```
thirdwing: The only change in this PR is the mojo runtime version update.

osery: :+1:.

thirdwing:
The `release-0.0.9` branch has been created: https://github.com/h2oai/dai-deployment-templates/commit/6f936e2612b1c78dc8f26bdfc9ceefe73c9a80bb

osery:
@thirdwing great, thanks.

osery:
I have created the same release on GitHub just to be consistent:
https://github.com/h2oai/dai-deployment-templates/releases/tag/v0.0.9

PR #63, opened 2019-05-20
Initial commit for SW MOJO scorer

g-eoj:
CSV formatting needs work.

### DIFF
--- a/sparkling-water-mojo-scorer/README.md
+++ b/sparkling-water-mojo-scorer/README.md
@@ -1,0 +1,32 @@
+# Example of using MOJO2 from PySparkling
+Multi-threaded MOJO scorer using Sparkling Water.
+
+## Requirements
+
+  - Java 8
+  - Configure `SPARK_HOME` to point to Spark distribution directory
+  - Configure `SPARKLING_HOME` to point to to Sparkling Water distribution directory
+  - Configure `DRIVERLESS_AI_LICENSE_FILE` to point to an active DAI license.sig file
+
+## (optional) 
+
+  - Configure `MASTER` to point to the master url for the cluster
+  - Create and use a Conda environment
+
+    ```
+    conda create -n mojo python=3.6
+    conda activate mojo
+    conda install openjdk
+    conda install numpy
+    conda install pandas
+    pip install h2o_pysparkling_2.4
+    ```
+
+## Usage
+
+Run `./run-example.sh mojo2-runtime.jar --mojo mp1/pipeline.mojo --input example.csv`
+
+## FAQ
+  - I have problems with showing Spark DataFrames due to ascii encoding error in Python session.
+    Please export, `export PYTHONIOENCODING=utf-8`
+
--- a/sparkling-water-mojo-scorer/example.py
+++ b/sparkling-water-mojo-scorer/example.py
@@ -1,0 +1,48 @@
+from pyspark.sql import SparkSession
+import argparse
+
+from pyspark.sql.types import StringType,FloatType
+from pyspark.sql.functions import udf
+
+def main():
+    spark = SparkSession.builder.appName('MOJO Pipeline Transformer').getOrCreate()
+    spark.sparkContext.setLogLevel('WARN')
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--mojo", type=str, required=True, help="Location of .mojo file, e.g., 'file:///tmp/pipeline.mojo'")
+    parser.add_argument("--input", type=str, required=True, help="Location of input file, e.g., 'file:///tmp/example.csv'")
+    parser.add_argument("--output", type=str, required=False, help="Location of output file, if not specified output is printed to console")
+    parser.add_argument("--format", type=str, required=False, default='csv', help="Format of output file, if not specified format is csv")
+    parser.add_argument("--nthreads", type=int, required=False, default=1, help="Number of threads to use, defaults to 1")
+
+    args = parser.parse_args()
+
+    # Load MOJO2 pipeline
+    from pysparkling.ml import H2OMOJOPipelineModel
+    mojo = H2OMOJOPipelineModel.create_from_mojo(args.mojo)
+    mojo.set_named_mojo_output_columns(True)
+
+    print("=============== In/Out ==============")
+    print("Inputs : {}".format(mojo.get_input_names()))
+    print("Outputs: {}".format(mojo.get_output_names()))
+    print("=====================================")
+
+    # Load data
+    example_df = spark.read.csv(args.input, header=True).repartition(args.nthreads)
+    prediction_df = mojo.transform(example_df)
+
+    # Flatenize prediction column to allow export to file/console
+    if args.format == 'parquet':
+        flat_fce = lambda x: str(x)
+    else:
+        flat_fce = lambda x: str(x) # TODO
+    flat_udf = udf(flat_fce, StringType())
+    output_df = prediction_df.withColumn('prediction', flat_udf(prediction_df.prediction))
+    if args.output:
+        output_df.write.format(args.format).mode('overwrite').option("header", "true").save(args.output)
+        print("Predictions written to: {}".format(args.output))
+    else:
+        output_df.show(10, False)
+
+if __name__ == '__main__':
+    main()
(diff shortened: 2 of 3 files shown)


sparkling-water-mojo-scorer/example.py
```diff
@@ -0,0 +1,48 @@
...
+    mojo = H2OMOJOPipelineModel.create_from_mojo(args.mojo)
+    mojo.set_named_mojo_output_columns(True)
+
+    print("=============== In/Out ==============")
+    print("Inputs : {}".format(mojo.get_input_names()))
+    print("Outputs: {}".format(mojo.get_output_names()))
+    print("=====================================")
+
+    # Load data
+    example_df = spark.read.csv(args.input, header=True).repartition(args.nthreads)
```
nkpng2k: `spark.read.csv` wont work if user inputs parquet file I think. If we are to have output as possibly parquet then probably we should expect input to be parquet as well

osery:
Hi @g-eoj, I may need a bit more context for this. Is there a tracking bug? What is the intended use and audience?

The thing is that this repo is ideally meant for simple unopinionated yet production ready templates that can be run by DAI (or some other automation) to deploy trained models to production. We had a discussion a while back if educational examples should be part of this repo and decided against that. I think we should have another repo for examples with the primary goal of educational clarity as opposed to production readiness. I believe that these are two rather conflicting goals. Can you help me understand where this code belongs to?

We can also talk over slack in my evening (>10pm CEST) to clarify this.

sparkling-water-mojo-scorer/run-example.sh
```diff
@@ -0,0 +1,49 @@
...
+  SPARK_HOME     : ${SPARK_HOME}
+  SPARKLING_HOME : ${SPARKLING_HOME}
+  SPARKLING_WATER_ZIP: ${SPARKLING_WATER_ZIP}
+  LAUNCHER_CMD   : ${CMD}
+  APP PARAMS     : $@
+  JAVA           : $(java -version)
+
+EOF
+
+$CMD example.py $@
```
osery: nit: or maybe `"$@"` just in case?

sparkling-water-mojo-scorer/example.py
```diff
@@ -0,0 +1,49 @@
...
+    mojo = H2OMOJOPipelineModel.create_from_mojo(args.mojo)
+    mojo.set_named_mojo_output_columns(True)
+
+    print("=============== In/Out ==============")
+    print("Inputs : {}".format(mojo.get_input_names()))
+    print("Outputs: {}".format(mojo.get_output_names()))
+    print("=====================================")
+
+    # Load data
+    example_df = spark.read.\
```
osery: nit: Please consider avoiding the ending `\`. One option is to enclose the expression in parentheses.

sparkling-water-mojo-scorer/example.py
```diff
@@ -0,0 +1,49 @@
...
+
+    args = parser.parse_args()
+
+    # Load MOJO2 pipeline
+    from pysparkling.ml import H2OMOJOPipelineModel
+    mojo = H2OMOJOPipelineModel.create_from_mojo(args.mojo)
+    mojo.set_named_mojo_output_columns(True)
+
+    print("=============== In/Out ==============")
+    print("Inputs : {}".format(mojo.get_input_names()))
```
osery: nit: I would tend to use format strings instead in the new code for the added clarity.
E.g., `print(f"Inputs : {mojo.get_input_names()}")`

sparkling-water-mojo-scorer/example.py
```diff
@@ -0,0 +1,49 @@
...
+import argparse
+
+from pyspark.sql.types import StringType,FloatType
+from pyspark.sql.functions import monotonically_increasing_id
+
+def main():
+    spark = SparkSession.builder.appName('MOJO Pipeline Transformer').getOrCreate()
+    spark.sparkContext.setLogLevel('WARN')
+
+    parser = argparse.ArgumentParser()
```
osery: clarity: Consider moving this into a dedicated function, e.g., `_parse_args` to let `main` focus on the important stuff.
Also please consider keeping line length under 80 (or at least 100). It makes github PR diff much more readable :+1:.

sparkling-water-mojo-scorer/example.py
```diff
@@ -0,0 +1,49 @@
...
+    parser.add_argument("--mojo", type=str, required=True, help="Location of .mojo file, e.g., 'file:///tmp/pipeline.mojo'")
+    parser.add_argument("--input", type=str, required=True, help="Location of input file, e.g., 'file:///tmp/example.csv'")
+    parser.add_argument("--output", type=str, required=False, help="Directory location of output files, if not specified output is printed to console")
+    parser.add_argument("--format", type=str, required=False, default='csv', help="Format of output file, if not specified format is csv")
+    parser.add_argument("--format_options", type=dict, required=False, default={'header': True}, help="Input options for the underlying data source")
+    parser.add_argument("--nthreads", type=int, required=False, default=1, help="Number of threads to use, defaults to 1")
+
+    args = parser.parse_args()
+
+    # Load MOJO2 pipeline
```
osery: style: Could we extract the independent logic to helper functions? Basically any of these line comments is trying to say that the following block is an unrelated responsibility. I would personally tend to have helper functions: `_load_mojo_pipeline(path)`, `_load_input_data(...)`, `_write_output(path)` or something similar. Note how the line comments become an actual code as they get transformed into function names (this increases their chance of being kept up to date as opposed to the inline comments that tend to get stale and eroded).

sparkling-water-mojo-scorer/README.md
```diff
@@ -0,0 +1,32 @@
...
+  - Configure `SPARKLING_HOME` to point to to Sparkling Water distribution directory
+  - Configure `DRIVERLESS_AI_LICENSE_FILE` to point to an active DAI license.sig file
+
+## (optional) 
+
+  - Configure `MASTER` to point to the master url for the cluster
+  - Create and use a Conda environment
+
+    ```
+    conda create -n mojo python=3.6
```
nkpng2k: do we want to be dependent on recommending `conda`. If anything, I would suggest shipping a template dockerfile.

sparkling-water-mojo-scorer/example.py
nkpng2k: agree. Alternatively, use `Click` module. Its a bit more legible and slightly easier to use imho

sparkling-water-mojo-scorer/example.py
```diff
@@ -0,0 +1,49 @@
+from pyspark.sql import SparkSession
+import argparse
+
+from pyspark.sql.types import StringType,FloatType
+from pyspark.sql.functions import monotonically_increasing_id
+
+def main():
```
nkpng2k: I might consider making this a class `SparklingWaterMojoScorer` or something. That way it is able to be passed around a little easier. Also, if its a bit more abstract users would be able to override certain functions for their own preferences?

sparkling-water-mojo-scorer/example.py
nkpng2k: agree with @osery

g-eoj:
The README still needs work but I think I addressed all the other comments.

osery:
Hi @g-eoj we still need to resolve the major comment above about whether this belongs here at all. @mmalohlava can you help deciding this?

> Hi @g-eoj, I may need a bit more context for this. Is there a tracking bug? What is the intended use and audience?
> 
> The thing is that this repo is ideally meant for simple unopinionated yet production ready templates that can be run by DAI (or some other automation) to deploy trained models to production. We had a discussion a while back if educational examples should be part of this repo and decided against that. I think we should have another repo for examples with the primary goal of educational clarity as opposed to production readiness. I believe that these are two rather conflicting goals. Can you help me understand where this code belongs to?
> 
> We can also talk over slack in my evening (>10pm CEST) to clarify this.

g-eoj:
@osery sorry I didn't respond to that message sooner. The motivation comes from multiple support requests for a method to do local multi-threaded predictions with the mojo scorer. @mmalohlava suggested we add the solution to this repo.

osery:
Thanks for the explanation @g-eoj, just spoke with @mmalohlava and we decided to create another repo for the examples (goal: simple, clear, self-contained): https://github.com/h2oai/dai-deployment-examples.
I have a feeling, this one would fit better there. Wdyt? (my apologies for the work involved)

PR #86, opened 2019-08-02
Keep build forever upon a successful deployment to S3

achraf-mer:


Jenkinsfile
```diff
@@ -100,6 +100,14 @@ pipeline {
                     }
                 }
             }
+            post {
```
michal-raska: I would treat Jenkins jobs as volatile, therefore I would not rely on Jenkins as any kind of "history keeper". I think git tags might be useful as well in this case. But otherwise looks good!

osery: We have branches and github releases too. I am ok with this.

nit: I would tend to define a function for this so that the high level pipeline reads a bit easier:
```
post {
    success {
        keepBuildLog()
    }
}
```

mmalohlava: @michal-raska i need to pin the job to be able to trace from release to a job which built the release. This just simplify work and allow to answer question like "what was status of CI tests when you release X.Y.Z".

PR #87, opened 2019-08-06
Upgrade mojo2-runtime to version 2.1.1

mmalohlava:
To support 1.6.x DAI.


mmalohlava:
thx! The MOJO upgrade should be safe, since the runtime should be backward compatible. However, still need to verify it.

PR #97, opened 2019-08-26
Update deployment template version to 0.2.7

nkalonia1:


osery:
Just came across this one by an accident. Please note that there is a step-by-step guide for bumping the version of the templates here: https://github.com/h2oai/dai-deployment-templates#step-by-step.
Let me know if it needs any clarifications.

Only mentioning this to prevent having `master` without the `-SNAPSHOT` prefix. This should be better automatized... :(

nkalonia1:
Thanks for the guide! Closing because the version was already increased.

PR #98, opened 2019-08-28
Extract sample request building code to a dedicated class.

osery:
Moves the exiting code and adds tests.
This is a step towards adding a dedicated endpoint which would return a sample scoring request discussed in h2oai/h2oai-serving#251.

### DIFF
--- a/aws-lambda-scorer/lambda-template/src/main/java/ai/h2o/mojos/deploy/aws/lambda/MojoScorer.java
+++ b/aws-lambda-scorer/lambda-template/src/main/java/ai/h2o/mojos/deploy/aws/lambda/MojoScorer.java
@@ -3,7 +3,11 @@ package ai.h2o.mojos.deploy.aws.lambda;
 
 import ai.h2o.mojos.deploy.common.rest.model.ScoreRequest;
 import ai.h2o.mojos.deploy.common.rest.model.ScoreResponse;
-import ai.h2o.mojos.deploy.common.transform.*;
+import ai.h2o.mojos.deploy.common.transform.MojoFrameToResponseConverter;
+import ai.h2o.mojos.deploy.common.transform.RequestChecker;
+import ai.h2o.mojos.deploy.common.transform.RequestToMojoFrameConverter;
+import ai.h2o.mojos.deploy.common.transform.SampleRequestBuilder;
+import ai.h2o.mojos.deploy.common.transform.ScoreRequestFormatException;
 import ai.h2o.mojos.runtime.MojoPipeline;
 import ai.h2o.mojos.runtime.frame.MojoFrame;
 import ai.h2o.mojos.runtime.lic.LicenseException;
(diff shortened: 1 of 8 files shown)


aws-lambda-scorer/lambda-template/src/main/java/ai/h2o/mojos/deploy/aws/lambda/MojoScorer.java
```diff
@@ -3,10 +3,7 @@
 
 import ai.h2o.mojos.deploy.common.rest.model.ScoreRequest;
 import ai.h2o.mojos.deploy.common.rest.model.ScoreResponse;
-import ai.h2o.mojos.deploy.common.transform.MojoFrameToResponseConverter;
-import ai.h2o.mojos.deploy.common.transform.RequestChecker;
-import ai.h2o.mojos.deploy.common.transform.RequestToMojoFrameConverter;
-import ai.h2o.mojos.deploy.common.transform.ScoreRequestFormatException;
+import ai.h2o.mojos.deploy.common.transform.*;
```
zoido: no

Don't use wildcard imports please. 

https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/SampleRequestBuilder.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * SampleRequestBuilder builds sample requests that pass all request validation and get actually scored.
+ * The resulting score is likely to be useless. The purpose is to give the caller an example request to further play
+ * with and fill with actual meaningful data.
+ */
+public class SampleRequestBuilder {
+    /**
+     * Builds a valid {@link ScoreRequest} based on the given mojo input {@link MojoFrameMeta}.
+     */
+    public ScoreRequest build(MojoFrameMeta inputMeta) {
```
zoido: can be static? if so can the class be static? if so can we have a private constructor?

aws-lambda-scorer/lambda-template/src/main/java/ai/h2o/mojos/deploy/aws/lambda/MojoScorer.java
osery: Let me blame this on my IntelliJ settings ;o). Fixed.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/SampleRequestBuilder.java
osery: Could be, but I quite like how I can test this separately and mock it in the `RequestCheckerTest`. This would not be possible with a static method & class. Wdyt?

zoido: There were times when I would insist... :)

osery:
Apologies @orendain. I got carried away and merged this before you had a chance to look at it. I am happy to implement any comments in a follow up PR. Sorry for that.

orendain:
> Apologies @orendain. I got carried away and merged this before you had a chance to look at it. I am happy to implement any comments in a follow up PR. Sorry for that.

No worries!  LGTM

PR #103, opened 2019-08-29
Use google-java-format formatting.

osery:
Run google-java-format (https://github.com/google/google-java-format) on the (non-KDB) code base.
Should only be whitespace and import reordering changes.

Will enable automated check as part of gradle build in a separate PR.

Why Google format?
 - It is not too permissive giving us the chance to focus on what really matters: code.
 - Supported in IDEs and build systems (mvn, gradle)
 - Me and @zoido are quite used to it already.
 - I quite like it :).

### DIFF
--- a/aws-lambda-scorer/lambda-template/src/main/java/ai/h2o/mojos/deploy/aws/lambda/MojoScorer.java
+++ b/aws-lambda-scorer/lambda-template/src/main/java/ai/h2o/mojos/deploy/aws/lambda/MojoScorer.java
@@ -1,4 +1,3 @@
-
 package ai.h2o.mojos.deploy.aws.lambda;
 
 import ai.h2o.mojos.deploy.common.rest.model.ScoreRequest;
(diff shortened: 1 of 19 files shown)


common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/MojoPipelineToModelInfoConverterTest.java
```diff
@@ -1,5 +1,7 @@
 package ai.h2o.mojos.deploy.common.transform;
```
zoido: Please double-check this file. The diff seems a bit weird.

osery: Thanks. I checked and it seems just a diff glitch, the actual differences are really just whitespace and import/accessibility order :+1:.

aws-lambda-scorer/lambda-template/src/main/java/ai/h2o/mojos/deploy/aws/lambda/MojoScorer.java
```diff
@@ -18,7 +17,6 @@
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3ClientBuilder;
 import com.amazonaws.services.s3.model.S3Object;
-
 import java.io.IOException;
```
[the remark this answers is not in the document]
osery: Yeah, I wouldn't say it is in any way useful tbh, other than being fixed and enforced.
I don't believe anyone "reads" the imports for fun :smile:.

PR #104, opened 2019-08-29
Fix google java style violations.

osery:
Fixes violations reported by checkstyle with google Java style configuration:
- https://google.github.io/styleguide/javaguide.html
- https://checkstyle.sourceforge.io/google_style.html

Will enable automated check using checkstyle as part of gradle build in a separate PR.

Why Google style?
- It is not too permissive giving us the chance to focus on what really matters: code.
- Me and @zoido are quite used to it already.
- I quite like it :).

### DIFF
--- a/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/CsvToMojoFrameConverter.java
+++ b/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/CsvToMojoFrameConverter.java
@@ -9,7 +9,7 @@ import java.io.InputStream;
 
 /** Converts the CSV from {@link InputStream} to the {@link MojoFrame} for scoring. */
 public class CsvToMojoFrameConverter {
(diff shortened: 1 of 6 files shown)


common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/CsvToMojoFrameConverter.java
```diff
@@ -9,6 +9,7 @@
 
 /** Converts the CSV from {@link InputStream} to the {@link MojoFrame} for scoring. */
 public class CsvToMojoFrameConverter {
+  /** Convert the CSV data from the given {@link InputStream} to a {@link MojoFrame}. */
```
zoido: I cannot find the reference but I am 99.9% sure that summaries should not be in the imperative from. So in this case it should be `Converts`.

same for `MojoScorer.score`

osery: Thanks, done.

PR #106, opened 2019-08-30
Restructure Gradle files

osery:
This is a follow up on @orendain comments in: https://github.com/h2oai/dai-deployment-templates/pull/105.

Turns out some of the comments cannot be implemented without having a lot of code copy&pasted around. E.g. `plugins` cannot be used anywhere except as the first part of a (sub)project gradle file. I want to prevent copying the shared configuration around as it is then hard to keep in sync. The result is that I still use the old way of enabling plugins where needed.

However the comments lead me to try a different way of structuring the gradle files. Please have a look and comment heavily. Maybe you won't find it better, in which case I am happy to drop the PR.

Things I personally like now are:
- it is possible to look at the subproject gradle file and see where stuff is defined (as the `apply from:` now work as sort of includes).
- separate stuff is in separate files, making the root gradle file much simpler.
- no need for central KDB related exclusions as this is now achieved by the kdb subprojects not using the style mixins (explicitely via `aplyt from: ... _no_style.gradle`).

Thoughts?

### DIFF
--- a/common/transform/build.gradle
+++ b/common/transform/build.gradle
@@ -1,13 +1,15 @@
+apply from: project(":").file('gradle/java.gradle')
+
 dependencies {
-    compile project(':common:rest-java-model')
+    implementation project(':common:rest-java-model')
 
-    compile group: 'ai.h2o', name: 'mojo2-runtime-api'
-    compile group: 'ai.h2o', name: 'mojo2-runtime-impl'
+    implementation group: 'ai.h2o', name: 'mojo2-runtime-api'
+    implementation group: 'ai.h2o', name: 'mojo2-runtime-impl'
 
-    testCompile group: 'com.google.truth.extensions', name: 'truth-java8-extension'
-    testCompile group: 'org.mockito', name: 'mockito-core'
-    testCompile group: 'org.mockito', name: 'mockito-junit-jupiter'
-    testCompile group: 'org.junit.jupiter', name: 'junit-jupiter-api'
+    testImplementation group: 'com.google.truth.extensions', name: 'truth-java8-extension'
+    testImplementation group: 'org.mockito', name: 'mockito-core'
+    testImplementation group: 'org.mockito', name: 'mockito-junit-jupiter'
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api'
     testRuntime group: 'org.junit.jupiter', name: 'junit-jupiter-engine'
 }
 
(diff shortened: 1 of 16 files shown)


osery:
Reverted the swagger part as it was incorrect and it turns out that there is not enough shared stuff to warrant for extraction into a dedicated mixin file. Sorry for the noise.

gradle/mixins/checkstyle.gradle
```diff
@@ -0,0 +1,9 @@
+// Defines shared Gradle project Checkstyle analyzer configuration.
+
+apply plugin: 'checkstyle'
+
+checkstyle {
+    toolVersion '8.21'
```
orendain: Guessing it is better practice to set the version here instead of `gradle.properties` because it defines the tool version and not the plugin version?  I'm just looking to learn, no issues whatsoever with this.

common/transform/build.gradle
```diff
@@ -1,13 +1,15 @@
+apply from: project(":").file('gradle/java.gradle')
+
 dependencies {
-    compile project(':common:rest-java-model')
+    implementation project(':common:rest-java-model')
 
-    compile group: 'ai.h2o', name: 'mojo2-runtime-api'
-    compile group: 'ai.h2o', name: 'mojo2-runtime-impl'
+    implementation group: 'ai.h2o', name: 'mojo2-runtime-api'
+    implementation group: 'ai.h2o', name: 'mojo2-runtime-impl'
```
orendain: ultra nit: There is a newline between `implementation group:...` and `testImplementation group:...` blocks here, while other build files (kbd, local-rest-scorer) do not.

I have no issue with this whatsoever, just pointing out in case you prefer one way or another.

gradle/mixins/checkstyle.gradle
osery: Ha, good catch. Tbh, I didn't really think about it. Now that you pointed that out, `gradle.properties` seems a better place for consistency with the other version information. Moved. Thank you :pray:.

common/transform/build.gradle
osery: Thanks @orendain. So I unified this now as follows:
- non-test deps
- newline
- test deps
Tbh, I don't care too much as long as the different types are not mixed. If you prefer some other way, I am happy to do that (I understand it is not really important, but it does not really cost anything either, so why not :)).

osery:
Will merge now, but happy to fine tune as we go in separate PRs.

PR #110, opened 2019-09-02
Switch docker building to Jib (Google Container Tools).

osery:
This has been suggested by @zoido to be able to build without a docker daemon and a Dockerfile.

Part of h2oai/h2oai-serving#262.


osery:
Changed the build task dependence to `jibBuildTar` instead of `jibDockerBuild`. This target does almost everything, except pushing to local docker, so it can be executed on machines without docker. Hope it is ok. But feel free to shout and I will implement comments as a separate PR.

PR #111, opened 2019-09-02
Push docker images to harbor on master and release branches.

osery:
Enable pushing docker images to Harbor for `master` and `release-*` branches or when explicitly requested by setting the PUSH_DOCKER_IMAGES pipeline parameter.

Part of https://github.com/h2oai/h2oai-serving/issues/262.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,9 +22,17 @@ pipeline {
         buildDiscarder(logRotator(numToKeepStr: '10'))
     }
 
+    parameters {
+        booleanParam(
+            name: 'PUSH_DOCKER_IMAGES',
+            defaultValue: false,
+            description: 'Whether to also push Docker images to Harbor.',
+        )
+    }
+
     stages {
 
-        stage('Init') {
+        stage('1. Init') {
             agent { label NODE_LABEL }
             steps {
                 script {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -22,9 +22,17 @@ pipeline {
...
+            name: 'PUSH_DOCKER_IMAGES',
+            defaultValue: false,
+            description: 'Whether to also push Docker images to Harbor.',
+        )
+    }
+
     stages {
 
-        stage('Init') {
+        stage('1. Init') {
```
michal-raska: just a note: I find the stage numbers a little bit inelegant. If there are expected checks enabled on GitHub and for some reason you need to add a new stage in the middle of the list, you need to change numbers on all previous stages and you also need to change the required checks. But that might be irrelevant in case of this repository.

osery: Great comment. You are totally right @michal-raska :+1:.

I only do this to have the stages in GitHub displayed in the expected execution order. So far it didn't bother us too much. As we look at these in GH way more often than edit these. As soon as this becomes painful I am all for reverting. If you know of another way to ensure the ordering that would be great too :pray:.

osery: Enable pushing docker images to Harbor for `master` and `release-*` branches or when explicitly requested by setting the PUSH_DOCKER_IMAGES pipeline parameter.

Part of https://github.com/h2oai/h2oai-serving/issues/262.

PR #118, opened 2019-09-18
Add mojo python cpp deployment template

achraf-mer:


aws-lambda-scorer-python/lambda-template/py_scorer.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class NotFound(ScorerError):
+    """Not found"""
+    status_code = 404
+
+
+def get_object_from_s3(bucket_name, bucket_key):
+    try:
+        s3.Bucket(bucket_name).download_file(bucket_key, '/tmp/pipeline.mojo')
```
achraf-mer: I think we should create a memory reader for the cpp runtime to not have to rely on storing mojo on disk

achraf-mer:
Quick note: it is a must that both java and cpp wrappers have the same http contract (request/response).

aws-lambda-scorer-python/lambda-template/py_scorer.py
thirdwing: That can be added. Can you open an issue in mojo2 repo?

I will work on this after we release 1.7.2.

achraf-mer: https://github.com/h2oai/mojo2/issues/736

aws-lambda-scorer-python/terraform-recipe/api_gateway.tf
```diff
@@ -0,0 +1,75 @@
+resource "aws_api_gateway_rest_api" "scorer_api" {
```
osery: Let's find a way to reuse these instead of copy pasting this from the Java part. Now if we need to change anything, we have to go to two places and keep them in sync. Also it is not clear what the actual difference is this way. Seems to me it is only the `handler` and `runtime`.
Maybe we could use the existing terraform recipe and turn it into a module: https://blog.gruntwork.io/how-to-create-reusable-infrastructure-with-terraform-modules-25526d65f73d.

aws-lambda-scorer-python/lambda-template/py_scorer.py
```diff
@@ -0,0 +1,120 @@
+import os
+import json
+import boto3
+import botocore
+import daimojo.model
+import datatable as dt
+
+s3 = boto3.resource('s3')
```
osery: Let's not have a global. Globals are evil, especially is their creation is not side effect free and can fail. Also we surely want a test (which is hard to do cleanly with globals). I would prefer wrapping the stuff here into an object that get's s3 upon construction. I guess we will need a global on the top-most level to represent the scorer object, but that is acceptable (as there is no way around it). Globals on the module level are imho not.

That way, the test would be relatively simple with a mock. Also, as an object, we can more easily download the mojo only once and reuse it.

aws-lambda-scorer-python/lambda-template/py_scorer.py
```diff
@@ -0,0 +1,120 @@
...
+    if body is None or len(body.keys()) == 0:
+        raise BadRequest("Invalid request. Need a request body.")
+
+    if 'fields' not in body.keys() or not isinstance(body['fields'], list):
+        raise BadRequest("Cannot determine the request column fields")
+
+    if 'rows' not in body.keys() or not isinstance(body['rows'], list):
+        raise BadRequest("Cannot determine the request rows")
+
+    get_object_from_s3(os.environ.get('DEPLOYMENT_S3_BUCKET_NAME'), os.environ.get('MOJO_S3_OBJECT_KEY'))
```
osery: Surely, we don't want to download the mojo each time a simple scoring request is received. Let's turn this into an object (cover it with tests) and have it download the mojo and create the model only once.

aws-lambda-scorer-python/lambda-template/py_scorer.py
```diff
@@ -0,0 +1,120 @@
...
+    m = daimojo.model("/tmp/pipeline.mojo")
+
+    if all(feature in body.keys() for feature in m.feature_names):
```
osery: Let's move the request validation into a separate well-named function for better readability. :+1:

aws-lambda-scorer-python/lambda-template/py_scorer.py
```diff
@@ -0,0 +1,120 @@
...
+        raise BadRequest("Bad request body")
+
+    columns_length = len(body['fields'])
```
osery: Let's move the request parsing into a separate well-named function for readability.

aws-lambda-scorer-python/lambda-template/py_scorer.py
```diff
@@ -0,0 +1,120 @@
...
+            'body': json.dumps(score(json.loads(event['body']), request_id))
+        }
+    except ScorerError as e:
+        return {
+            'isBase64Encoded': False,
+            'statusCode': e.status_code,
+            'body': json.dumps({'id': request_id, 'message': str(e)})
+        }
+    except Exception as exc:
+        print(exc)
```
osery: Please consider using the `logging` library and add a textual message too.

aws-lambda-scorer-python/lambda-template/py_scorer.py
```diff
@@ -0,0 +1,120 @@
...
+    d_frame = dt.fread(
+        text=tmp_frame.to_csv(),
+        columns=list(types.keys()),
+        na_strings=m.missing_values
+    )
+
+    res = m.predict(d_frame)
+
+    return {
+        'id': request_id,
```
osery: I would tend to leave the result formatting on the `request_handler` function. It is already formatting the result and handling errors, so let's keep it localized.

aws-lambda-scorer-python/lambda-template/py_scorer.py
```diff
@@ -0,0 +1,120 @@
...
+        'score': list(zip(*res.to_list()))
+    }
+
+
+def request_handler(event, context):
+    request_id = ''
+    try:
+        if 'requestContext' not in event.keys() or 'requestId' not in event['requestContext']:
+            raise BadRequest("Request does not seem to be relayed through an AWS API Gateway")
+        else:
```
osery: nit: No `else:` needed here.

aws-lambda-scorer-python/lambda-template/py_scorer.py
```diff
@@ -0,0 +1,120 @@
...
+        if event['resource'] != '/score':
+            raise NotFound("Resource not found.")
+
+        if 'body' not in event.keys():
+            raise BadRequest("Invalid request. Need a request body.")
+
+        return {
+            'isBase64Encoded': False,
+            'statusCode': 200,
+            'body': json.dumps(score(json.loads(event['body']), request_id))
```
osery: I would tend to call the function outside the return statement. This way, it is quite hard to spot even though it is the most important stuff in this function.

aws-lambda-scorer-python/lambda-template/Makefile
```diff
@@ -0,0 +1,25 @@
...
+	@echo "  help           display this help"
+	@echo "  clean          clean existing artifacts"
+	@echo "  package        create a lambda deployment package named lambda_function.zip"
+	@echo
+
+clean:
+	rm -rf aws_lambda_env
+	rm -rf pip_package
+
+package: clean
```
osery: I am not sure how we are going to integrate this with the main gradle build system of the repo.
Currently we build stuff on pure java jdk docker image where this would clearly not work :o(.

achraf-mer: Agree, we might need a different system other than jdk specific

aws-lambda-scorer-python/lambda-template/py_scorer.py
achraf-mer: Not sure if Lambda serves traffic from one machine, in my experience most cdn don't, we might still need to re-download the file on every request or at least check that we have it on disk.

aws-lambda-scorer-python/lambda-template/py_scorer.py
achraf-mer: This goes along with my previous comment.

aws-lambda-scorer-python/terraform-recipe/api_gateway.tf
achraf-mer: Planned for next iteration. Will need to change code on the other template to include a variables.tf, this would mean some changes to the ci as well.

aws-lambda-scorer-python/lambda-template/py_scorer.py
osery: Multiple requests run in quick succession are likely to hit an existing instance. AWS will spawn multiple when traffic is high and shut them down if it is low. Every now and then, we will have to reinitialize, but we should not do that on every request.

Please don't implement this by checking the disk file. We really want an entity that is initialized once and then reused rather than external contract about disk file being present. Note that this is a public repo and we want to keep it clean and tidy :+1:.

aws-lambda-scorer-python/terraform-recipe/api_gateway.tf
osery: It's fine (or even preferred) to do it separately, but let's do it before this one is merged then. I would rather not have copy&pasted code around unless strictly needed. Is that ok?

aws-lambda-scorer-python/lambda-template/Makefile
james94: PyGradle could be a potential solution to integrating this project with the main gradle build system of the repo. PyGradle leverages Gradle and is an enterprise Python build system. You could look at switching Makefile with Pygradle.
https://github.com/linkedin/pygradle

PR #122, opened 2019-10-01
issue-121: Initial layout of DAI Deployment Templates doc

ABartzGit:
Note that Sphinx cannot render markdown when the path to those markdown files is backward. The example README in this is a direct copy of the aws-lambda-scorer/README. It renders correctly. But going backward to different folders does not work. (Which is why we stick with .rst files)


osery:
@angela0xdata what is the reasoning behind having a separate generated documentation for this repo? I understand that the DAI part is already covered here: https://github.com/h2oai/h2oai/blob/dev/docs/deployment.rst. Where would be the generated documentation placed/used and what would be the target audience?

I admit that unless there is a clear usecase, I would personally prefer having the documentation placed along side the code in the way that is native (and common) in GitHub public repos, i.e., readme files placed in the corresponding folders (don't care much whether markdown or rst). This way, we don't have to handle versioning of the documentation as it is always there with the git commit and anybody can view it on GitHub.

Please give me a bit more context and reasoning behind this. Thank you :+1:.

osery:
Also please don't commit the `.DS_Store` files.

osery:
One more idea that crossed my mind, would it be possible to use Gradle plugin instead of make?
There is at least: https://trustin.github.io/sphinx-gradle-plugin/.
The goal would be to use something that can integrate with the current build system of the repo (which is Gradle). Never tried it, just wondering if that would be feasible.

docs/conf_works.py
```diff
@@ -0,0 +1,348 @@
+# -*- coding: utf-8 -*-
```
osery: What is the purpose of this file? Is it needed and how is it triggered?
Certainly the copyright seems a bit off :-/.

docs/sphinxext/docscrape.py
```diff
@@ -0,0 +1,497 @@
...
+
+"""
+
+import inspect
+import textwrap
+import re
+import pydoc
+from StringIO import StringIO
+from warnings import warn
+4
```
osery: What is the `4` here? This cannot possibly work :thinking:, do we even use it?

angela0xdata: I don't know. I didn't create this file :)

docs/conf_works.py
angela0xdata: Sorry, I was testing two different conf.py files. I'll delete this and ignore the ds_store file

angela0xdata:
@osery The reasoning behind this PR is that I was asked to create this. Also, I have no preference over whether docs are built with make or gradle.

docs/sphinxext/docscrape.py
osery: Who did then :)? Can we remove it?
I would rather not commit broken stuff into the repo (not that this one is public and thus I want it to be as clean as possible).

docs/conf_works.py
osery: Thanks :+1:.

osery:
@angela0xdata would you be willing to try to find the answers for me (since I don't know who else to ask)?
Many thanks :pray:.

angela0xdata:
@osery sure, I'll do my best! :)

g-eoj:
per @mmalohlava request we should try to document mojo scoring with sparkling water on spectrum conductor.

osery:
@g-eoj I think there is a better place for that in the https://github.com/h2oai/dai-deployment-examples repo. An example of a very nice doc describing the use of the python scoring pipeline (credits to @mtanco ) is here: https://github.com/h2oai/dai-deployment-examples/blob/master/python-scoring-pipeline/http_ec2_ubuntu.md. I would personally tend to go that way.

g-eoj:
per @mmalohlava request putting this here: https://www.ibm.com/developerworks/community/blogs/281605c9-7369-46dc-ad03-70d9ad377480/entry/Submitting_Batch_Spark_Applications_to_IBM_Platform_Conductor?lang=en

PR #123, opened 2019-10-02
AzureML Template

JHerna17:
Deployment Template for Azure ML Services. 

Contains examples for Local, Azure Container Instances and AKS.


mmalohlava:
@JHerna17 we need to improve this PR. 
We need production ready template, no example.

osery:
Maybe we can start by putting this notebook into the examples repo to have something public.

I am not at all familiar with AzureML. What does it take? A docker image? Should we reuse the Java REST scorer (to keep the API) and make changes needed to make AzureML happy?

JHerna17:
> 
> 
> Maybe we can start by putting this notebook into the examples repo to have something public.
> 
> I am not at all familiar with AzureML. What does it take? A docker image? Should we reuse the Java REST scorer (to keep the API) and make changes needed to make AzureML happy?

It only needs a Machine Learning workspace (can be created via script), a enviroment yaml file (azure creates one for you, but you can also define it) and a python file with the init and run function defined. 

I will modify the PR to be a deployment ready script with 2 options: Deploy to ACI or Deploy to AKS.

mmalohlava:
Actions:
- [ ] push updated scripts
- [ ] explore local deployment and testing
- [ ] update documentation with more details about running the script
- [ ] use our Swagger API
- [ ] describe test scenarios

azure-ml-deployment/dai_deploy.yml
```diff
@@ -0,0 +1,9 @@
+name: dai_deploy
+dependencies:
+  - python=3.6.2
+  - pip:
+    - azureml-defaults
+    - numpy
+    - pandas
+    - https://h2ostore.blob.core.windows.net/marketplacescripts/AzureDeployments/daimojo-2.0.1%2Bmaster.478-cp36-cp36m-linux_x86_64.whl
```
mmalohlava: Needs to be parameterized.

azure-ml-deployment/score.py
```diff
@@ -0,0 +1,32 @@
...
+    
+    os.environ["DRIVERLESS_AI_LICENSE_KEY"] = dai_license    
+    
+    import daimojo.model
+
+    global model
+    model = daimojo.model("./scoring-pipeline/pipeline.mojo")
+    
+
+def run(raw_data):
```
mmalohlava: What is in `raw_data`?

azure-ml-deployment/scoring-pipeline/dai_deploy.yml
```diff
@@ -0,0 +1,9 @@
+name: dai_deploy
+dependencies:
+  - python=3.6.2
+  - pip:
+    - azureml-defaults
+    - numpy
+    - pandas
+    - https://h2ostore.blob.core.windows.net/marketplacescripts/AzureDeployments/daimojo-2.0.1%2Bmaster.478-cp36-cp36m-linux_x86_64.whl
```
mmalohlava: needs parameterization.

azure-ml-deployment/test.py
```diff
@@ -0,0 +1,25 @@
+import requests
+import json
+import pandas as pd 
+
+service_uri = "http://eab43710-5b28-4228-ae7b-df236b23e6b2.eastus.azurecontainer.io/score"
+service_key = "27gZi21I5em5T5sJ4qwpcaTgwIYnhqYq"
+
```
achraf-mer: Please remove the info above. This repo is public.

azure-ml-deployment/DeployModel.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
```
achraf-mer: It would be better to converge to using one common deployment method/mechanism. For instance, the Lambda Java uses gradle, Lambda Python uses a Makefile, and this Azure uses a shell script.

JHerna17:
Latest commits with the changes from the reviews.

azure-ml-deployment/DeployModel.sh
JHerna17: Right now the only way to do the deployment is using the Azure CLI.

nkpng2k: https://docs.microsoft.com/en-us/azure/machine-learning/service/tutorial-deploy-models-with-aml

is there any way we can implement via python script then? As it is, this does not really mesh nicely with our ultimate goal which is to have 1-click deployment in dai no?

azure-ml-deployment/deploymentconfigAKS.json
```diff
@@ -0,0 +1,21 @@
...
+    },
+    "dataCollection":
+    {
+        "storageEnabled": true
+    },
+    "authEnabled": true,
+    "containerResourceRequirements":
+    {
+        "cpu": 0.5,
+        "memoryInGB": 1.0
```
nkpng2k: what is the maximum here? 1 GB? that seems really small considering...

azure-ml-deployment/dai_deploy.yml
```diff
@@ -0,0 +1,9 @@
+name: dai_deploy
+dependencies:
+  - python=3.6.2
+  - pip:
+    - azureml-defaults
+    - numpy
+    - pandas
+    - https://h2ostore12792.blob.core.windows.net/mojodependencies/daimojo-2.0.1%2Bmaster.478-cp36-cp36m-linux_x86_64.whl
```
nkpng2k: these should probably point to the released artifacts in s3 imho

JHerna17: I'm going to remove this file since it gets created every time the script is run. 
The script takes the path for the daimojo and datatable packages and upload them to a blob storage under the user subscription.

azure-ml-deployment/deploymentconfigAKS.json
JHerna17: This example is indeed a deployment config for K8s on Azure (aka AKS) these files are meant to be modified by the user.

azure-ml-deployment/DeployModel.sh
achraf-mer: It should not matter what it uses under the hood as long as we create a common interfaces for all of our deployments. Ultimately we will need to deploy from a python script, so agree with Nick, a python wrapper would be nice.

nkpng2k: I think python would be idea... we will really not want to add dependency on azure cli to dai

azure-ml-deployment/Python/config.txt
```diff
@@ -0,0 +1,15 @@
+workspace_name="DAI_DEMO2"
+resource_group_name="DAI_DEPLOY2"
+location="eastus"
+deployment_type="ACI"
+service_name="mydemoserviceaci"
+aks_cluster="myaks"
+model_name="MyDaiModel"
+description="Credit Card Demo"
+model_path="../mojo/pipeline.mojo"
+daimojo_path="../scoring-pipeline/env/dependencies/daimojo-2.0.1+master.478-cp36-cp36m-linux_x86_64.whl"
```
mmalohlava: is this example config? if yes, then should be in "example" folder, if it is config template, there should be placeholders instead of the hard-coded values.

azure-ml-deployment/Python/DeployModel.py
```diff
@@ -0,0 +1,188 @@
...
+from random import randint
+#Azure ML
+from azureml.core import Workspace
+from azureml.core.model import Model
+from azureml.core.model import InferenceConfig
+from azureml.core.webservice import AciWebservice, AksWebservice, Webservice
+from azureml.core.compute import AksCompute, ComputeTarget
+from azureml.core.webservice import LocalWebservice
+#Azure Blob
+from azure.storage.blob import BlockBlobService, PublicAccess
```
nkpng2k: just a warning... but this came up in DAI issues with dependencies.... looks like the above module BlockBlobService is now considered legacy.

JHerna17: Updated to the latest version of the Python SDK (v12).

PR #124, opened 2019-10-04
Remove model ID from the endpoint paths.

osery:
Our deployment templates only support loading a single model and we don't plan to change that.
As a result, having the model id in the request path is only a nuisance that makes calling the score unnecessarily complicated.

This PR changes that so that before too many people start relying on the current convoluted request paths.

Bumps version to `1.0.0-SNAPSHOT` to make the incompatible API change more obvious.

### DIFF
--- a/common/swagger/swagger.yaml
+++ b/common/swagger/swagger.yaml
@@ -144,8 +144,6 @@ definitions:
   ScoreRequest:
     type: object
     properties:
-      id:
-        type: string
       includeFieldsInOutput:
         type: array
         items:
--- a/common/swagger/swagger.yaml
+++ b/common/swagger/swagger.yaml
@@ -25,7 +25,7 @@ paths:
       description: Returns unique id of the model loaded in the server and used for scoring
       operationId: getModelId
       produces:
-        - application/json
+        - text/plain
       parameters: []
       responses:
         '200':
--- a/common/swagger/swagger.yaml
+++ b/common/swagger/swagger.yaml
@@ -53,8 +53,8 @@ paths:
     get:
       tags:
         - metadata
-      summary: sample scoring request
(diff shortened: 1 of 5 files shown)


osery:
Thanks @nkpng2k. Lambda should not be affected. Is there anything else we should be concerned about?

nkpng2k:
> Thanks @nkpng2k. Lambda should not be affected. Is there anything else we should be concerned about?

@osery I think the only thing we might need to be concerned about is the deployment ui for dai... since I want to say that the endpoint examples are hardcoded...

common/swagger/swagger.yaml
```diff
@@ -17,126 +17,96 @@ schemes:
...
-        - models
-      summary: Returns a list of models loaded in the server
-      description: Returns a list of models loaded in the server
-      operationId: getModels
+        - metadata
+      summary: Returns model id
+      description: Returns unique id of the model loaded in the server and used for scoring
+      operationId: getModelId
       produces:
         - application/json
```
orendain: @osery Should this be text/plain since we're returning just the model-id string itself?

osery: Good point, thanks :pray:. Fixed.

osery:
@orendain you are totally right about the `id`, it seems to have absolutely no value (in the input request that is). Removed as well.

osery:
@nkpng2k good point! Let me:
- merge this one
- create a release of the templates
- send a PR updating templates in DAI and changing the fixed paths at the same time.
Thanks all :pray:.

osery:
All the steps in https://github.com/h2oai/dai-deployment-templates/pull/124#issuecomment-541647566 have been completed. Shout if there is any issue :pray:.

PR #127, opened 2019-10-15
Deduplicate docker pushing in Jenkinsfile.

osery:
I would like to have a helper for the push itself but the gradle parameters are a bit too different :-/.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,12 +24,12 @@ pipeline {
 
     parameters {
         booleanParam(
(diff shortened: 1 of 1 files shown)


osery:
The pushed public image for Deployer is visible here: https://hub.docker.com/r/h2oai/deployer/tags
Tags: `latest`, `f2d670bdfcce9c6a0116bae7235f66ac73f2c9b7`, and `0.1.0_f2d670b`.

PR #128, opened 2019-10-16
Add an optional field for specifying field name holding unique row ids.

osery:
Motivated by the discussion over the monitoring PR here:
http://github.com/h2oai/h2oai-monitor/pull/93

Please let me know if I misunderstood the original motivation and this is not helpful.

Also took this opportunity to add description to the individual request fields.

### DIFF
--- a/common/swagger/swagger.yaml
+++ b/common/swagger/swagger.yaml
@@ -185,7 +185,9 @@ definitions:
         description: >
           An array of rows consisting the actual scoring output. The order of rows corresponds to the order of the input
           request rows. Each row contains any copied input fields first (in the order of appearance in the input row).
-          If the `idField` was specified and also listed in the  but not provided it follows after the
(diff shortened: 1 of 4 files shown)


common/swagger/swagger.yaml
```diff
@@ -145,23 +145,47 @@ definitions:
...
     properties:
       id:
+        description: >
+          A unique id of the model used for scoring.
         type: string
       score:
+        description: >
+          An array of rows consisting the actual scoring output. The order of rows corresponds to the order of the input
+          request rows. Each row contains any copied input fields first (in the order of appearance in the input row).
+          If the `idField` was specified and also listed in the  but not provided it follows after the
```
orendain: The last line is unfinished.

osery: Good catch :+1:, fixed.

osery:
@orendain tbh, my decision about the ordering of output fields does look a bit unnatural in the retrospect :-(. I don't feel confident changing it now though as it would affect AWS lambda too.

PR #130, opened 2019-11-04
sagemaker template

g-eoj:
Requested by customer.


nkpng2k:
Probably, we will want to ensure that the build artifacts are not added to the distribution zip file. Only built and converted to docker image, since it is unlikely that we add this implementation to DAI.

@osery wdyt? I added the `distributionZip` task to build.gradle in the sagemaker project for convenience sake, but thinking about it, probably we don't need.

aws-sagemaker-hosted-scorer/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
```
nkpng2k: probably don't need this. Should be defined by top level project

aws-sagemaker-hosted-scorer/gradlew
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env sh
```
nkpng2k: this too...  Should be defined by top level project

aws-sagemaker-hosted-scorer/gradlew.bat
```diff
@@ -0,0 +1,84 @@
+@if "%DEBUG%" == "" @echo off
```
nkpng2k: also here,  Should be defined by top level project

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/Serve.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        long startMillis = System.currentTimeMillis();
+        model = MojoPipeline.loadFrom(f.getAbsolutePath());
+        long endMillis = System.currentTimeMillis();
+        long deltaMillis = endMillis - startMillis;
+        System.out.println("INFO: MojoPipeline.loadFrom took " + deltaMillis + " millis");
+    }
+
+    HashMap<String,String> invocations(HashMap<String,String> inputs) {
```
osery: I think we already have most of this code in the shared libraries. For example the LocalRestServer does not do much work itself: https://github.com/h2oai/dai-deployment-templates/blob/master/local-rest-scorer/src/main/java/ai/h2o/mojos/deploy/local/rest/controller/ModelsApiController.java.

Would it make sense to take a similar approach? The benefit I see is that we would have the same REST interface for free (at least the payloads) and have the request transformation already covered by tests. Wdyt?

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/Serve.java
```diff
@@ -0,0 +1,103 @@
...
+            rowBuilder.setValue(key, value);
+        }
+        frameBuilder.addRow(rowBuilder);
+        MojoFrame iframe = frameBuilder.toMojoFrame();
+
+        // Make the prediction.  Prediction outputs go to the oframe.
+        MojoFrame oframe = model.transform(iframe);
+
+        // Move the oframe fields into a more standard Java object for the response.
+        @SuppressWarnings("unchecked")
```
osery: Please don't suppress warnings ☮️. Instead, we probably want to fix the issue they point at.

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/Serve.java
```diff
@@ -0,0 +1,103 @@
...
+        for (int i = 0; i < outCsv.getNCols(); i++) {
+            String key = outCsv.getLabels()[i];
+            String value = outCsv.getData()[0][i];
+            response.put(key, value);
+        }
+
+        return response;
+    }
+
+    @SuppressWarnings("InfiniteLoopStatement")
```
osery: As noted above, this should not be here.

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/ServeController.java
```diff
@@ -0,0 +1,101 @@
...
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import com.google.gson.JsonParser;
+
+@SuppressWarnings("unused")
```
osery: I don't think these suppressions are needed. Please remove them.

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/ServeController.java
```diff
@@ -0,0 +1,101 @@
...
+@RestController
+@CrossOrigin
+public class ServeController {
```
osery: Ideally, I would tend to keep only Sage specific stuff in this class and reuse the shared code for the parsing and such. See: https://github.com/h2oai/dai-deployment-templates/blob/master/local-rest-scorer/src/main/java/ai/h2o/mojos/deploy/local/rest/controller/ModelsApiController.java

May you need some changes on the shared code portions, but we can always do that. Imho that would be better than having many copies of similar request parsing code with minor alterations. Wdyt?

aws-sagemaker-hosted-scorer/Sagemaker-DriverlessAI-MOJO.ipynb
```diff
@@ -0,0 +1,188 @@
+{
```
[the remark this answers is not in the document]
nkpng2k: good point. Its not required for execution. Just an example

aws-sagemaker-hosted-scorer/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+    }
+    to {
+        image = dockerRepositoryPrefix + 'sagemaker-hosted-scorer'
+        tags = [version]
+    }
+    container {
+        ports = ['8080']
+        volumes = [
+                // mojo pipeline and license file will live here
+                '/opt/ml/model'
```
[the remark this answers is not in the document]
nkpng2k: @osery ah yes... this is specific to sagemaker. They expect the models and licenses etc to go to a specific place. In the deploy process is very similar to the python notebook example. 
1. define the model that you want to use
2. define what the endpoint (deployment) should look like e.g. which image, vm size, etc
3. deploy. 

and sagemaker handles all the backend stuffs, getting mojo from s3 moving to their required location, etc.

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/Serve.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Autowired
+    public Serve() {
+    }
+
+    /**
+     * Call this immediately after the constructor above.
+     * Since this object is used as an Autowired Spring object, the constructor call is not actually visible.
+     */
+    void init() throws Exception {
```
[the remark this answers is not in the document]
nkpng2k: I expect that this is sagemaker specific stuffs, I believe they require an init method... Will have to double check

nkpng2k:
@osery thank you for the comprehensive review. I started going through it and then realized that a lot of the comments are related to the original work done by tomk. Will look through when I have the chance, need to understand the java better if I am to make changes. @g-eoj Idk if you are able to do this... or maybe prompt tom?

aws-sagemaker-hosted-scorer/example_request.py
```diff
@@ -0,0 +1,53 @@
+import boto3
```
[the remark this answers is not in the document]
nkpng2k: tests would be nice... will work on that.

aws-sagemaker-hosted-scorer/build.gradle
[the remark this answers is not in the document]
nkpng2k: as per sagemaker docs:
```
In your CreateModel request, the container definition includes the ModelDataUrl parameter, which identifies the S3 location where model artifacts are stored. Amazon SageMaker uses this information to determine where to copy the model artifacts from. It copies the artifacts to the /opt/ml/model directory for use by your inference code.

```

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/ServeController.java
```diff
@@ -0,0 +1,101 @@
...
+            e.printStackTrace();
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+            return null;
+        }
+
+        try {
+            return serve.invocations(inputs);
+        }
+        catch (Exception e) {
+            e.printStackTrace();
```
mmalohlava: what is this? 😱

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/Serve.java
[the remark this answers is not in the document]
osery: I doubt sagemaker sees into the code that deeply tbh 🤔 . My guess is that this is part of the hack about not serving requests before the mojo is loaded, which I hope can be better solved by using Spring native means (described in the other comments).

aws-sagemaker-hosted-scorer/build.gradle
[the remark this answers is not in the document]
osery: Thanks for the detailed explanation, makes sense. Til 👍.

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/Serve.java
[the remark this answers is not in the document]
nkpng2k: yes this was my mistake.... responded before fully understanding the code myself.

nkpng2k:
@osery this should be very close to completed. reused MojoScorer/moved. Only worry is that ModelApiController is now a lot of copy paste, but should be fine, user can still make requests to all the request mappings despite sagemaker only requiring `/invocations` and `/ping`

Just a thought here. Idk, maybe it makes sense to have the `ModelApi` abstract class define fewer requestMappings, and then define any additional mappings in the lower down subproject. WDYT?

aws-sagemaker-hosted-scorer/examples/Sagemaker-DriverlessAI-MOJO.ipynb
```diff
@@ -0,0 +1,188 @@
+{
```
osery: qq: Would it make sense to mention this in the readme?

aws-sagemaker-hosted-scorer/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+```
+tar cvf mojo.tar pipeline.mojo license.sig
+gzip mojo.tar
+aws s3 cp mojo.tar.gz s3://<your-bucket>/
+```
+
+Next create the appropriate model and endpoint on Sagemaker.
+Check that the endpoint is available with `aws sagemaker list-endpoints`.
+See `example_request.py` for example of end point query.
```
osery: Now moved to `examples/example_request.py`, right?

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/config/ScorerConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @Bean
+  public MojoScorer mojoScorer(
+      RequestToMojoFrameConverter requestConverter,
+      MojoFrameToResponseConverter responseConverter,
+      MojoPipelineToModelInfoConverter modelInfoConverter,
+      CsvToMojoFrameConverter csvConverter) {
+    return new MojoScorer(requestConverter, responseConverter, modelInfoConverter, csvConverter);
+  }
+}
```
osery: ultra nit: New line at EOF please (just to remove the visual clutter here and limit uninteresting diff in the future) :pray:.

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/controller/ModelsApiController.java
```diff
@@ -0,0 +1,108 @@
...
+     * {@link MojoScorer} initialized class containing loaded mojo, and mojo interaction methods
+     * @param sampleRequestBuilder
+     * {@link SampleRequestBuilder} Simple class for generating sample request.
+     */
+    @Autowired
+    public ModelsApiController(MojoScorer scorer, SampleRequestBuilder sampleRequestBuilder) {
+        this.scorer = scorer;
+        log.info("Mojo pipeline successfully loaded ({}).", scorer.getModelId());
+        this.sampleRequestBuilder = sampleRequestBuilder;
+        this.isReady = true;
```
osery: This can never be `false` after c-tor, right? And Spring would surely not call it before c-tor is done. So I think the entire variable is not very useful. Personally, I would tend to just remove it.

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/controller/ModelsApiController.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    @Override
+    public ResponseEntity<ScoreRequest> getSampleRequest() {
+        return ResponseEntity.ok(sampleRequestBuilder.build(scorer.getPipeline().getInputMeta()));
+    }
+
+    @RequestMapping("/ping")
+    public ResponseEntity<String> ping() {
```
osery: This seems to always return `ok`. I would tend to remove both the field `isReady` and the getter `getReady` and only return a constant `ok` here.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -116,7 +113,7 @@ private static MojoPipeline loadMojoPipelineFromFile() {
         !Strings.isNullOrEmpty(MOJO_PIPELINE_PATH),
         "Path to mojo pipeline not specified, set the %s property.",
         MOJO_PIPELINE_PATH_PROPERTY);
-    log.info("Loading Mojo pipeline from path {}", MOJO_PIPELINE_PATH);
+    //    log.info("Loading Mojo pipeline from path {}", MOJO_PIPELINE_PATH);
```
osery: Why do we remove the logger? Is it because of Sage?

nkpng2k:
> I agree that unifying the shared code would be a nice improvement. Please, feel free to give it a go. Could be also a separate PR or an issue to track that. Your call.

yes, I think let us track that in a separate pr/issue. So that we can press forward with this for customer.

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/controller/ModelsApiController.java
nkpng2k: yea... i just borrowed a version of this from tomk code.... will think on this because in production, sagemaker makes periodic checks to `/ping` to verify whether or not the scorer is alive. Would want something that ends up throwing an error if it fails.

osery: Having a `/ping` endpoint is ok, but as long as we always return `true` (which is the case, right?), I would remove the extra unused logic. We can put it back once we know how to actually decide between being healthy or not. Does it make sense or am I missing something 🤔? Not a big deal though :+1:.

nkpng2k: ahhh i see what you mean, if ever we are able to respond from `/ping`, the server is already alive anyways. should just return true

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
nkpng2k: oops was supposed to add the logger back... thanks for pointing this out!

aws-sagemaker-hosted-scorer/src/main/java/ai/h2o/mojos/deploy/sagemaker/hosted/controller/ModelsApiController.java
nkpng2k: :+1:

aws-sagemaker-hosted-scorer/build.gradle
```diff
@@ -0,0 +1,61 @@
+plugins {
+    id 'java'
+    id 'idea'
+    id 'org.springframework.boot'
+    id 'io.spring.dependency-management'
+    id 'com.google.cloud.tools.jib'
+}
+apply from: project(":").file('gradle/java_no_style.gradle')
```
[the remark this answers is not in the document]
nkpng2k: I had thought I had changed that, thanks for the reminder!

nkpng2k:
ok @osery I'm inclined to say this is good to go... would appreciate one more quick pass by you, whenever you have the chance. Thanks!

aws-sagemaker-hosted-scorer/examples/example_request.py
```diff
@@ -0,0 +1,32 @@
+import boto3
+
+client = boto3.client('sagemaker-runtime')
+
+endpoint_name = "endpoint-name"
+content_type = "application/json"
+payload = """
```
osery: Optional: We could also use the `json` library here to have a native Python struct converted to Json (see here: https://docs.python.org/3/library/json.html). Feel free to keep. Your call.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -116,7 +129,7 @@ private static MojoPipeline loadMojoPipelineFromFile() {
         !Strings.isNullOrEmpty(MOJO_PIPELINE_PATH),
         "Path to mojo pipeline not specified, set the %s property.",
         MOJO_PIPELINE_PATH_PROPERTY);
-    log.info("Loading Mojo pipeline from path {}", MOJO_PIPELINE_PATH);
+        log.info("Loading Mojo pipeline from path {}", MOJO_PIPELINE_PATH);
```
osery: nit: Wrong indent.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -130,13 +143,10 @@ private static MojoPipeline loadMojoPipelineFromFile() {
     }
     try {
       MojoPipeline mojoPipeline = MojoPipeline.loadFrom(mojoFile.getPath());
```
osery: nit: Should we keep the other log entries here too? Not pushing, just asking.

nkpng2k:
@osery seems like a broke this by adding back logging... looks like the jar fails to launch with error about too many log4j configs. Maybe you have some thoughts? 

@g-eoj can you possibly provide the traceback... Or I can replicate later. 

Thanks!

osery:
@nkpng2k it seems like we are getting LogBack dependency transitively. I personally don't care either way (LogBack vs. Log4j). We can do something like: https://github.com/h2oai/dai-deployment-templates/pull/134 which should leave the actual logging backend unspecified. Tried the local rest scorer and it seems to work. Wdyt?

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
nkpng2k: I believe I moved the log line into the `ModelsApiController.java` so its not gone... just moved. 

but this is the better location. less code

nkpng2k:
container runs locally after final commits, and now passes jenkins linting tests. Merging as no changes to logic have been made since approval.

PR #134, opened 2019-11-19
Only depend on the slf4j api in the shared code.

osery:
This is to leave the actually logging back-end choice on the leaf binary package.


nkpng2k:
@osery I think this looks good. Will test locally, it its happy, can I merge it into my branch?

osery:
Sure thing 👍.

nkpng2k:
Side note, it looks like linting is failing because of the spacing in the sagemaker project, will fix.

PR #137, opened 2019-11-20
use ${version} instead of version in Jenkinsfile

thirdwing:


Jenkinsfile
```diff
@@ -112,7 +112,7 @@ pipeline {
                     s3upDocker {
                         localArtifact = "build/dai-deployment-templates-${version}.zip"
                         artifactId = 'dai-deployment-templates'
-                        version = version
+                        version = ${version}
```
michal-raska: it needs to be a string like `version = "${version}"`.

thirdwing: Just added quote

PR #138, opened 2019-11-20
New release after jenkinsfile fix

thirdwing:
Both of 1.0.1 and 1.0.2 release have been uploaded into wrong places


mmalohlava:
in theory we do not need a new releaes, we just need to correct structure in S3.

PR #140, opened 2019-11-21
Mention bucket_name in README.

osery:
This should make the README up-to-date again.

### DIFF
--- a/aws-lambda-scorer/README.md
+++ b/aws-lambda-scorer/README.md
@@ -27,7 +27,8 @@ e.g., for testing purposes.
 
 ### One-off Setup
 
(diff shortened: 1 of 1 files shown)


surenH2oai:
Terraform is not supported on PPC https://github.com/hashicorp/terraform/pull/18708
We will not be able to support Lambda deployment of DAI models form PPC.   

Our document is accurate to let user know deploying to lambda is not supported on PPC.
http://docs.h2o.ai/driverless-ai/latest-lts/docs/userguide/deployment.html


The best option is to document manual steps to deploy the model/runtime (MOJO/Java) on AWS.

@tkmaker Please advise will documentation be OK for PPC?  or is it worth automating the deployment on PPC to Lambda as a script?

PR #143, opened 2019-12-09
Return field names in the scoring response by default.

osery:
This can be disabled by passing in `"noFieldNames": true` in the scoring request and works both for AWS lambda and local rest scorer.

This is a relatively safe option as it adds a new field that can be easily ignored by existing clients. There is still a small chance of breaking an existing client who might have strict JSON parsing enabled. E.g., for Java using `FAIL_ON_UNKNOWN_PROPERTIES` (https://fasterxml.github.io/jackson-databind/javadoc/2.6/com/fasterxml/jackson/databind/DeserializationFeature.html#FAIL_ON_UNKNOWN_PROPERTIES)

**Q:** Why not build flag or java property?
**A:** These would apply globally for all requests. I think it is better when we give users the option to switch between more and less verbose on the fly as needed. This allows low-bandwidth when working with production queries as well as human debugging, when needed, without restarting the scorer.

### DIFF
--- a/aws-lambda-scorer/README.md
+++ b/aws-lambda-scorer/README.md
@@ -111,60 +111,6 @@ The expected response should follow this structure, but the actual values may di
 ```json
 {
   "id": "a12e7390-b8ac-406a-ade9-0d5ea4b63ea9",
-  "score": [
-    [
-      "1.0",
-      "0.6240277982943945",
-      "0.045458571508101536",
-      "0.330513630197504"
-    ],
-    [
-      "3.0",
-      "0.7209441819603676",
-      "0.06299909138586585",
-      "0.21605672665376663"
-    ],
-    [
-      "4.0",
-      "0.7209441819603676",
-      "0.06299909138586585",
-      "0.21605672665376663"
-    ]
-  ]
-}
-```
-
-To get back also the names of fields in the order in which they appear in the resulting `score`,
-pass in an extra request field `includeFieldNames` set to `true`.
-
-```json
-{
-  "fields": [
-    "sepal_len", "sepal_wid", "petal_len", "petal_wid"
-  ],
-  "includeFieldsInOutput": [
-    "sepal_len"
-  ],
-  "includeFieldNames": true,
-  "rows": [
-    [
-      "1.0", "1.0", "2.2", "3.5"
-    ],
-    [
-      "3.0", "10.0", "2.2", "3.5"
-    ],
-    [
-      "4.0", "100.0", "2.2", "3.5"
-    ]
-  ]
-}
-```
-
-Resulting in a more verbose answer good for manual queries and debugging:
-
-```json
-{
-  "id": "a12e7390-b8ac-406a-ade9-0d5ea4b63ea9",
   "fields": [
     "sepal_len",
     "class.Iris-setosa",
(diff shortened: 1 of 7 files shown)


aws-lambda-scorer/README.md
```diff
@@ -110,6 +110,67 @@ The expected response should follow this structure, but the actual values may di
...
+      "0.7209441819603676",
+      "0.06299909138586585",
+      "0.21605672665376663"
+    ]
+  ]
+}
+```
+
+To get back also the names of fields in the order in which they appear in the resulting `score`,
+pass in an extra request field `includeFieldNames` set to `true`.
```
mmalohlava: Note: should it be called `includeClassNamesInOutput`?

osery: The way it is implemented, it turns names for all the elements in the output. This includes both classes and any fields copied from the input, plus optionally the is field. Wdyt?

nkpng2k: I'm tempted to say that this should be `true` by default, and `false` only by request... would invite less confusion to first time users, and realistically should not be costly enough where it ever needs to be ignored except in special cases...

tomkraljevic:
i would also say enable by default.

osery:
Changed to make returning the fields the default. Inverted and renamed the flag as "noFieldNames" so that `false` and missing the flag have the same behavior (both are JSON falsy).

I kept the name derived for "field" (as opposed to "class"). This is consistent with the input request field `fields` (which I don't dare changing). Moreover, it also contains possibly more values than resulting class names (e.g., names of fields copied from the input and possibly also the `idField` name).

PTAL.

aws-lambda-scorer/README.md
osery: Thanks @nkpng2k, done.

nikhilgk: > Note: should it be called `includeClassNamesInOutput`?
> 
Since this change was done, shouldn't the JSON response field name be "class" or "targetClass" instead of "fields"?

osery: I implemented the content regarding the default value and tried to explain why the "class" weird does not match. Please see my other comment below :+1:.

aws-lambda-scorer/README.md
```diff
@@ -110,6 +110,13 @@ The expected response should follow this structure, but the actual values may di
 
 ```json
 {
+  "id": "a12e7390-b8ac-406a-ade9-0d5ea4b63ea9",
+  "fields": [
```
nikhilgk: Should we change this to "class" or "classNames" or "targetClasses"?

tomkraljevic:
the specific word “class” is a perhaps a poor choice since it is a reserved keyword in many languages, including java.

aws-lambda-scorer/README.md
osery: Please read my other comments. This is not only classes and there is also consistency with the input field.

nkpng2k: agree with @osery here... `fields` encompasses all header values returned, not just the mojo generated classes from target.

osery:
Discussed with @mmalohlava offline and decided to also rename `noFieldNames` to `noFieldNamesInOutput` for consistency with the `includeFieldsInOutput`.

osery:
I don't see any blocking comments, so merging now.

PR #149, opened 2020-01-10
Implement scorer to SQL database using JDBC

nkpng2k:
Attempt to create rest scorer that loads mojo + some conf file for configuring jdbc connection. And then scores data that results from SQL query and writes it back to the configured database

@osery probably the branch is in a good place for iteration. Thanks in advance for taking time to review when you have a chance. No massive hurry. Fairly certain this will not be packaged into the distributionZip file that gets implemented in dai, which is on purpose for now. 

@mmalohlava we can maybe use this instead of eric one in dai-deployment-examples moving forward? :smile:


common/jdbc/build.gradle
```diff
@@ -0,0 +1,25 @@
...
+    implementation group: 'org.apache.spark', name: 'spark-mllib_2.11'
+    implementation group: 'com.typesafe', name:'config'
+
+    zinc 'com.typesafe.zinc:zinc:0.3.9'
+    zinc 'org.scala-lang:scala-library:2.10.5'
+}
+
+test {
+    useJUnitPlatform()
+}
```
osery: nit: newline at eof pretty please (less diff noise).

nkpng2k:
Hi @osery. Thank you for the comment and concerns. From testing perspective, I can add a few unit tests and I have the scala styling tests that I've used in the past as well. Apologies, I kind of just assumed you had knowledge of scala :smile:

High level, the motivation was that we have an implementation of sql database scorer already in examples, but @mmalohlava have voiced multiple concerns about it, from usability/maintenance perspective. My understanding of this repository is that it is an aggregation of templates for mojo deployment that we are more "confident" in supporting/suggest as viable means of deployment. Hence the pr here. 

The plus side of this implementation, is that is more or less an extension of our jdbc connector in DAI already but with write capabilities and scoring from Kuba SW transform api. Spark handles the distribution of data, so theoretically a small implementation (little memory) of this scorer can score large batches of data... given enough time. Random, BigQuery has a jdbc driver so it should work with this scorer as well.

osery:
I see. Fine with me. Thanks for the explanation.

nkpng2k:
side note @osery maybe we can make the api slightly more flexible by having smaller yaml files? like one that defines `/model/id` and `/model/schema` only that everything can inherit, then we override the `/model/score` in independent yaml.... Idk if this is possible but would make things slightly less weighty.

common/jdbc/build.gradle
nkpng2k: whoops. okie will do

nkpng2k:
@mn-mikke Michal suggested maybe you should review the scala code here? No hurry, but when you have the opportunity. Thanks in advance!

common/jdbc/src/main/scala/ai/h2o/mojos/deploy/common/jdbc/SqlScorerClient.scala
```diff
@@ -0,0 +1,223 @@
...
+  def init(): SqlScorerClient = {
+    val sqlScorerClient = new SqlScorerClient()
+    sqlScorerClient
+  }
+}
+
+class SqlScorerClient {
+  val logger: Logger = LoggerFactory.getLogger(classOf[SqlScorerClient])
+
+  private final val conf: SparkConf = new SparkConf()
```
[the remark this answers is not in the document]
nkpng2k: thanks! I think you are right `val` is more than enough.

common/jdbc/src/main/scala/ai/h2o/mojos/deploy/common/jdbc/SqlScorerClient.scala
```diff
@@ -0,0 +1,223 @@
...
+      case SaveMethodEnum.OVERWRITE => SaveMode.Overwrite
+      case SaveMethodEnum.IGNORE => SaveMode.Ignore
+      case SaveMethodEnum.ERROR => SaveMode.ErrorIfExists
+      // Fall through case catches SaveMethodEnum.PREVIEW since it should never get passed here anyways
+      case _ => SaveMode.Append
+    }
+  }
+
+  private def castDataFrameToArray(df: DataFrame): Array[String] = {
+    df.limit(5).collect().map(_.toString())
```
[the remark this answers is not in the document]
nkpng2k: design decision :man_shrugging: Since we are writing the resultant score to the database, realistically it is a little wasteful to send the entire scored dataframe back to the user. So the user gets in the response up to 5 rows.

common/jdbc/src/main/scala/ai/h2o/mojos/deploy/common/jdbc/SqlScorerClient.scala
```diff
@@ -0,0 +1,223 @@
...
+  private def setSqlProperties(): Properties = {
+    val properties: Properties = new Properties()
+    properties.setProperty("user", jdbcConfig.dbUser)
+    properties.setProperty("password", jdbcConfig.dbPassword)
+    properties.setProperty("driver", jdbcConfig.dbDriver)
+    properties
+  }
+
+  private def castToInt(number: Any): Int = {
+    logger.info("Converting scala numeric types to java types")
```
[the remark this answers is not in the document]
nkpng2k: you are right... more like just casting to `Int`, changed the message.

common/jdbc/src/main/scala/ai/h2o/mojos/deploy/common/jdbc/SqlScorerClient.scala
```diff
@@ -0,0 +1,223 @@
...
+      case _ => SaveMode.Append
+    }
+  }
+
+  private def castDataFrameToArray(df: DataFrame): Array[String] = {
+    df.limit(5).collect().map(_.toString())
+  }
+
+  private def sanitizeInputString(input: String): String = {
+    val cleanedString: String = new String(input.getBytes(StandardCharsets.UTF_8))
```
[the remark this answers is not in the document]
nkpng2k: tbh, wasn't sure if http request had proper encoding... so decided to ensure it by recasting string with expected utf-8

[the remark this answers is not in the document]
mn-mikke: If you have a string in Java/Scala, isn't it always internally encoded as UTF-16? 

I would guess that the line converts UTF-16 string to UTF-8 byte array and then it uses the platform's default `CharsetDecoder` to convert byte array to to UTF-16 string. Is ti what you intended?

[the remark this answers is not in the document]
nkpng2k: I think you are right. I think this was originally put in when I was debugging some issue with string input but i think I remember the issue being not related to encoding. Probably I just need to take this line out

common/jdbc/src/main/scala/ai/h2o/mojos/deploy/common/jdbc/SqlScorerClient.scala
```diff
@@ -0,0 +1,192 @@
...
+    try {
+      new JdbcConfig(JDBC_CONFIG_FILE)
+    } catch {
+      case e: IOException =>
+        throw new RuntimeException("Unable to load JDBC Config file", e)
+    }
+  }
+
+  private def generatePartitionMapping(url: String, query: String, idColumn: String): Unit = {
+    val tmpDf: DataFrame = doQuery(url, query)
```
mn-mikke: Just a note: Most of Scala developers don't specify type of variables and private fields. It's definitely a matter of preferences, just mentioning it.

common/jdbc/src/main/scala/ai/h2o/mojos/deploy/common/jdbc/SqlScorerClient.scala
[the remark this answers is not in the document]
mn-mikke: I understand your motivation and it' absolutely valid. Just it will be quite confusing for people that will use this method in future. The name and parameters of the method are generic and the result will contain 5 records at most. If the constant was passed as an additional method parameter [here](https://github.com/h2oai/dai-deployment-templates/pull/149/files#diff-626a7507baa00a668a95d2aa9273aee8R62), the confusion would disappear IMHO. If this is just a PoC project, please ignore my comment.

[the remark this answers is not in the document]
nkpng2k: Thanks for the input here! I like your idea. Will make changes

common/jdbc/src/main/scala/ai/h2o/mojos/deploy/common/jdbc/SqlScorerClient.scala
nkpng2k: hmm yea just habitual from java I'd say... if its not overly terrifying, then I'd lean towards keeping the type defs and private fields, if only because I'm a little obsessive compulsive :smile:

common/jdbc/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api'
+    testRuntime group: 'org.junit.jupiter', name: 'junit-jupiter-engine'
+}
+
+test {
+    useJUnitPlatform()
+}
+
+bootJar {
+    enabled=false
```
mmalohlava: Why do you need `bootJar` section?

gradle/mixins/dependencies.gradle
```diff
@@ -35,5 +35,13 @@ dependencyManagement {
             entry 'slf4j-api'
             entry 'slf4j-log4j12'
         }
+        dependency group: 'org.apache.spark', name: 'spark-core_2.11', version: sparkVersion
+        dependency group: 'org.apache.spark', name: 'spark-sql_2.11', version: sparkVersion
+        dependency group: 'org.apache.spark', name: 'spark-mllib_2.11', version: sparkVersion
+        dependency group: 'org.scala-lang', name: 'scala-library', version: scalaVersion
+        dependency group: 'ai.h2o', name: 'sparkling-water-scoring_2.11', version: sparklingWaterVersion
+        dependency group: 'ai.h2o', name: 'sparkling-water-core_2.11', version: sparklingWaterVersion
```
mmalohlava: Do you need to have sparkling-water-core?

common/jdbc/build.gradle
nkpng2k: by default `bootJar` is set `enabled=true` and given that we need to add to the classpath for the scorer to work properly, this needed to be set to `false`

gradle/mixins/dependencies.gradle
nkpng2k: I feel like we did, but not sure, can check

nkpng2k: ya tested without sparkling water core. no issues. removed.

nkpng2k:
tested after rebase. good to go so merging. scoring is roughly .00033 seconds per row.

PR #157, opened 2020-02-12
Force Jib to create empty dirs for /mojos and /secrets.

osery:
These should be in the docker image to start with. Not having these and expecting inputs in them feels weird. Originally, I thought setting `volumes` was enough, but it turns out it is not.

Note that Jib copied anything in `./src/main/jib` to the root of the resulting docker image.

Needed to unblock @nkpng2k with #156.


osery:
Thanks @zoido, interesting. The `!.gitignore` seems superfluous, no? I find it a tidbit confusing as my goal is to keep it, not to ignore anything. After reading (scrolling really ;-)) through:
https://stackoverflow.com/questions/7229885/what-are-the-differences-between-gitignore-and-gitkeep, maybe the option I like the best is a README file, which as an added value, can describe what the dir is for. Thoughts?

zoido:
@osery README makes sense. Thanks.

osery:
Thanks. Filed #158 to rename these.

PR #166, opened 2020-04-03
[MOJO] upgrade mojo runtime to support xgboost 1.0.0

thirdwing:


thirdwing:
I also updated the `CsvToMojoFrameConverter` class with my own understanding.

Please directly update this branch if necessary. @mmalohlava @pkozelka

pkozelka:
@thirdwing I think your imlp is correct, so I just added a deprecation here, in order to redirect potential users to use the batched approach.

@mmalohlava is there any realistic chance that this class (`CsvToMojoFrameConverter`) is used by someone else ? It would be best to completely remove it and replace the call.

pkozelka:
@mmalohlava is this project working with large/huge datasets on its REST interface?
If so, refactoring it to streamful/batched approach would make quite a sense.
@g-eoj any thoughts?

thirdwing:
>@thirdwing I think your imlp is correct, so I just added a deprecation here, in order to redirect potential users to use the batched approach.

@pkozelka did you push your change? I didn't see it here.

pkozelka:
@thirdwing oh you are right I didn't ... and it seems that I am not permitted to push here, so the attempt to push failed and I didn't notice.
Here comes the patch, can you please review and add it? Thanks


`0001-deprecated-CsvToMojoFrameConverter-in-favor-of-Batch.patch`:

```
From 3cbfe2fbc738af9bd81aee3c6cbdb5c8469cd99d Mon Sep 17 00:00:00 2001
From: Petr Kozelka <pkozelka@gmail.com>
Date: Sat, 4 Apr 2020 10:19:10 +0200
Subject: [PATCH] deprecated CsvToMojoFrameConverter in favor of
 BatchedCsvMojoProcessor

---
 .../common/transform/CsvToMojoFrameConverter.java   | 13 +++++++++++--
 1 file changed, 11 insertions(+), 2 deletions(-)

diff --git a/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/CsvToMojoFrameConverter.java b/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/CsvToMojoFrameConverter.java
index c8acf05..4b39b61 100644
--- a/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/CsvToMojoFrameConverter.java
+++ b/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/CsvToMojoFrameConverter.java
@@ -12,9 +12,18 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.util.Iterator;
 
-/** Converts the CSV from {@link InputStream} to the {@link MojoFrame} for scoring. */
+/**
+ * Converts the CSV from {@link InputStream} to the {@link MojoFrame} for scoring.
+ * @deprecated use {@link BatchedCsvMojoProcessor} instead, it is better for large datasets as it can process it by batches of configurable size.
+ */
+@Deprecated
 public class CsvToMojoFrameConverter {
-  /** Converts the CSV data from the given {@link InputStream} to a {@link MojoFrame}. */
+  /**
+   * Converts the CSV data from the given {@link InputStream} to a {@link MojoFrame}.
+   * @deprecated use {@link BatchedCsvMojoProcessor} instead, it is better for large datasets as it can process it by batches of configurable size.
+   * An example usage is here: https://github.com/h2oai/mojo2/blob/master/java/mojo2-runtime-impl/src-examples/main/java/ai/h2o/mojos/ExecuteMojo.java#L67-L77 .
+   */
+  @Deprecated
   public MojoFrame apply(InputStream inputStream, MojoFrameBuilder frameBuilder)
       throws IOException {
     MojoRowBuilder rowBuilder = frameBuilder.getMojoRowBuilder();
-- 
2.17.1
```

PR #172, opened 2020-05-22
Remove useless override of one variant of method DummyPipeline.transform()

pkozelka:
... to prevent compatibility issues with planned changes in mojo2's MojoPipeline.

Tiny change. The overriden method which this PR removes has no functional impact (it's a `transform()`method that, in derived implementation, calls another variant of `transform()`). So it was always useless; however, now it might hurt, because this use prevents changing that method to final (only one single variant of transform will be overridable in future).


mmalohlava:
**⚠️ Note: pls do not merge yet - need to wait for upgrade of runtime. ⚠️**

PR #179, opened 2020-07-06
[BUILD] API safety check.

mmalohlava:
The change replace relative reference to Swagger API by module dependence.

It also introduces an API safety - if we change swagger file, the safety will force change of version.


gradle.properties
```diff
@@ -1,4 +1,15 @@
 version = 1.0.16-SNAPSHOT
+# This is version of swagger file common/swagger/swagger.yaml
+# If there is a minor change (e.g., adding a new method) in the swagger file we should bump minor version
+# If there is a major change (e.g., from v1 to v2) we should bump major version
+swaggerApiVersion = 1.0
+# This is safety hash of the swagger file to detect changes in the file.
+# If the change is detected, developer needs to manually update version and hash.
+swaggerApiHash = e92fde8f25d5ea4939a453bb14255cbd
+# Version: ${swaggerApiVersion}.patch
+patchVersion = 16-SNAPSHOT
```
nkpng2k, line 10: hmmm ok, so we expect that changes to the swagger api to drive the versioning rather than any changes to the java code.

I might say for proper versioning we should follow this: `<swagger api version>.<java changes>.<patch changes>`

meaning if we change the api, bump major version. If we add new java functionalities, bump minor version. Fixes == patch changes

mmalohlava: Good point! i think we do not need 2 version numbers to represent `swaggerApiVersion`. single major version is good enough.

jakubhava:
Closing as stale

PR #180, opened 2020-07-07
[ENHANCEMENT] example scoring script in python for local rest scorer

nkpng2k:
small python script example of how to score against local rest scorer with python `requests` package.

may help to resolve confusion in using the scorer and what the actual data payload should look like.


jakubhava:
Closing as stale

PR #181, opened 2020-07-08
[REST] Enable support of non-standard JSON tokens (NaN, Infinity, -Infinity)

mmalohlava:


local-rest-scorer/src/test/java/ai/h2o/mojos/deploy/local/rest/config/JsonConfigurationTest.java
```diff
@@ -0,0 +1,85 @@
...
+        Arguments.arguments("[-Infinity, -2, -1, 0.0]", toRow("-Infinity", "-2", "-1", "0.0"))
+    );
+  }
+
+  private static Row toRow(String... values) {
+    Row row = new Row();
+    row.addAll(Arrays.asList(values));
+    return row;
+  }
+}
```
nkpng2k, line 85: nit: new line :smile:

local-rest-scorer/src/main/java/ai/h2o/mojos/deploy/local/rest/config/JsonConfiguration.java
```diff
@@ -1,17 +1,20 @@
...
 
 @Configuration
 class JsonConfiguration {
 
   @Bean
   public ObjectMapper objectMapper() {
-    ObjectMapper mapper = new ObjectMapper();
-    mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
-    return mapper;
+    return new Jackson2ObjectMapperBuilder()
```
nkpng2k, line 15: qq - I believe this is the same for aws lambda as well... maybe needs change there also??

local-rest-scorer/src/test/java/ai/h2o/mojos/deploy/local/rest/config/JsonConfigurationTest.java
```diff
@@ -0,0 +1,85 @@
...
+  @ParameterizedTest
+  @MethodSource
+  public void defaultObjectMapperShouldFailOnNonNumericRows(String givenRowJson, Row expRow) throws IOException {
+    // When & Then
+    assertThrows(JsonMappingException.class, () -> defaultObjectMapper.readValue(givenRowJson, Row.class));
+
+  }
+
+  @SuppressWarnings("unused")
+  private static Stream<Arguments> objectMapperShouldHandlesMixedValueRows() {
```
nkpng2k, line 56: typo: objectMapperShouldHandleMixedValueRows

mmalohlava:
@orendain can this affect MLOps?

local-rest-scorer/src/main/java/ai/h2o/mojos/deploy/local/rest/config/JsonConfiguration.java
```diff
@@ -1,17 +1,20 @@
...
+        .serializationInclusion(JsonInclude.Include.NON_NULL)
+        .featuresToEnable(JsonParser.Feature.ALLOW_NON_NUMERIC_NUMBERS)
```
orendain, line 17: Since this is an asymmetric setting (i.e., this affects `JSON -> POJO` and not the other way around), I just want to confirm that the reverse was not intended (i.e., that models won't be returning "NaN"/"Inf" values while the column type is `(Float|Int)(32|64)`)?

Post-scoring, MLops doesn't currently expect non-numeric numbers, so it would require a few changes and perhaps some issues as our time series database (InfluxDB) doesn't support NaN/Inf as a numeric.

local-rest-scorer/src/main/java/ai/h2o/mojos/deploy/local/rest/config/JsonConfiguration.java
mmalohlava: Good point, but lambda is using gson (not the Spring objectmappers/jackson). But will need test as well.

local-rest-scorer/src/main/java/ai/h2o/mojos/deploy/local/rest/config/JsonConfiguration.java
mmalohlava: Good point, the intention is to handle only input (e.g., parsing JSON-serialized pandas frame which can contain non-standard JSON literal NaN). It should not affect output, however test for that (especially for returning NaN) is missing. so will be added.

btw: you have good point about NaN returning by a model. I would consider is as a bug (not enough attention payed to model corner cases), however, practically it could happen (e.g., if i give you a row with almost all NaNs, the result will be probably not defined...)

mmalohlava: @orendain i updated testing little bit - NaN are translated to nulls, but Float.Infinity is translated to "Infinity". 
It looks little bit inconsistent.

orendain: Thank you for the test cases @mmalohlava 🙏 !  I've opened an issue for Model Ops monitoring to handle the marshaling of these values ... Off the top of my head, I'm not sure how to best handle "Inf" values downstream (no support for it in InfluxDB, Grafana, downstream in drift detection [currently]) - but I'll give it some thought when tackling that and keep you in the loop!

mmalohlava: @orendain i would better to fix it on MOJO2 side - we do not need to provide -Inf/Inf, not sure if there is any benefit of producing such values.

Issue #129, opened 2019-10-29
[CHORE] Refactor Repo Structure to Allow Multiple Languages


nkpng2k:
@osery with cpp mojo's there will me more deployment strategies that implement python/alternative language to java.

Opening this issue to track thoughts here:
1. perhaps we can do something similar to what we have in DAI with a Makefile, and nest the projects into their own separate directories. e.g.
```
dai-deployment-templates
|-- Makefile
|-- python
        |-- cpp mojo project
        |-- azureml project
|-- java
        |-- original java mojo projects.
```
then we can change the java build process to dump the builds into a `build` directory... same as whatever we chose for the python side and then compile both build directories and zip them to ship. 

WDYT?

nkpng2k:
cc @mmalohlava @achraf-mer

osery:
Hi Nick, thanks for starting this.

Are there some alternatives? Personally, I find Make a very bad way to build projects. It does not really track dependencies so it is almost impossible to do correct incremental builds :(. The way DAI is build is imho super painful and you never know if the right thing gets actually built or not.

Gradle itself seems to have some support for Python and C++ builds:
- Python: https://github.com/linkedin/pygradle
- C++: https://docs.gradle.org/current/userguide/building_cpp_projects.html
I admit that I have never used these myself.

We use Bazel in Model Manager and build Golang and Python with it to our great happiness (it also supports Java and C++). Bazel has a rather steep learning curve, but once set up, it is really great way to build things quickly (remote caching) and consistently (tracks all dependencies).

I understand both these options are much more work, but I am really scared we could end up with thousands lines of make code that is super hard to maintain or even do correctly to start with. Wdyt? @mmalohlava

nkpng2k:
@osery thank you for pointing out bazel and pygradle. I would be inclined, actually, to lean towards one of those. Agree, the Makefile in DAI is super unwieldy.... Wouldn't mind being forced to learn bazel.

achraf-mer:
@nkpng2k not pushing to use a cli, just stating if something requires a cli, then the scope need to be limited, this goes for other cloud specific tools as well.

nkpng2k:
@achraf-mer can we maybe revisit this? I think given that now we have the c++ dai mojo, it is really time to start trying to make this repo multiple language supported. WDYT?

achraf-mer:
[...]

PR #194, opened 2020-09-08
implement MLflow Model wrappper for DAI MOJO

MartinBarus:
This PR implements simple command line tool that creates MLflow Model wrapper of Driverless AI MOJO model that can then be used by mlflow to be deployed in:

- Azure ML
- Amazon SageMaker
- Local REST Server
- Apache Spark UDF

https://www.mlflow.org/docs/latest/models.html


mlflow/example/license.sig
```diff
@@ -0,0 +1 @@
+ENTER VALID DRIVERLESS AI LICENSE
```
nkpng2k, line 1: nit: do we need this file at all? I would expect based on the scoring code below + readme that this should be understood, but I am fine with keeping it.

mlflow/model/dai_model.py
```diff
@@ -0,0 +1,106 @@
...
+    mlflow.pyfunc.save_model(
+        path=model_name,
+        python_model=DAIWrappedModel(),
+        artifacts=artifacts,
+        conda_env=conda_env,
+    )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
```
nkpng2k, line 81: nit: I'd generally advocate for making this a method, but that doesn't really change the functionality.

nkpng2k: or maybe just define full class with all the little bits and bobs in it :man_shrugging:

nkpng2k (approved):
@MartinBarus thanks for this PR. It does bring up a good question. Since this is primarily a java repo at the moment. I think it is a good time maybe to revisit this issue from some time back: https://github.com/h2oai/dai-deployment-templates/issues/129

I would maybe say for now to make the directory as `dai-deployment-templates/python/mlflow` simply to avoid confusion? wdyt?

jakubhava:
Closing as stale

PR #202, opened 2020-09-29
[MOJO] Update H2O-3 runtime to version 3.30.1.3.

orendain:
This updates H2O-3 runtime from `3.28.1.2` -> `3.30.1.3` (latest, at the time of writing) in order to support new H2O-3 MOJO versions (i.e. MOJO model version 1.40).

Doctor @mmalohlava: Mind if I open build/dev PRs to get a new release out?

> Edit: Modified target version to `3.30.1.2` as recommended.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,7 +4,7 @@ version = 1.0.24-SNAPSHOT
 # issue resolution.
 
 # Internal dependencies:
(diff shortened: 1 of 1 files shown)


mmalohlava (approved):
We have to figure out how to change it on a single place (only in mojo2) for example)

PR #229, opened 2021-04-29
Mo/gcp ai unified

mohamedasni:


gcp-unified-mojo-scorer/src/main/java/ai/h2o/mojos/deploy/gcp/unified/controller/ModelsApiController.java
```diff
@@ -0,0 +1,85 @@
...
+      log.debug(" - failure cause: ", e);
+      return ResponseEntity.badRequest().build();
+    }
+  }
+
+  // @Override
+  // public ResponseEntity<ScoreRequest> getSampleRequest() {
+  //   return ResponseEntity.ok(sampleRequestBuilder.build(scorer.getPipeline().getInputMeta()));
+  // }
+}
```
nkpng2k, line 85: nit: newline at end of files. Is nice for less diff in github.

common/gcp-unified/src/main/java/ai/h2o/mojos/deploy/common/transform/unified/CsvToMojoFrameConverter.java
```diff
@@ -0,0 +1,66 @@
+package ai.h2o.mojos.deploy.common.transform.unified;
```
nkpng2k, line 1: so this looks like an exact copy paste, is there a reason? like why not just use the actual one?

common/gcp-unified/src/main/java/ai/h2o/mojos/deploy/common/transform/unified/MojoFrameToResponseConverter.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.UUID;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Converts the resulting predicted {@link MojoFrame} into the API response object {@link
+ * ScoreResponse}.
+ */
+public class MojoFrameToResponseConverter
```
nkpng2k, line 26: so I just thought of something. Maybe it makes sense to do like a method override here, that way there is less copy paste. I believe it would be something like:
```
public class MyClass {
  public void doSomething(myString string) {}
  public void doSomething(myString, myInt int) {}
}
```
https://www.geeksforgeeks.org/overloading-in-java/ might be useful... idk I hate the idea of so much copy paste

common/gcp-unified/src/main/java/ai/h2o/mojos/deploy/common/transform/unified/MojoPipelineToModelInfoConverter.java
```diff
@@ -0,0 +1,45 @@
+package ai.h2o.mojos.deploy.common.transform.unified;
```
nkpng2k, line 1: tbh none of this should be in the `common` project. It seems (even if lots of copy paste) that it is specific to gcp unified. So it deserves to go in gcp unified project I think. Otherwise, other projects will end up having this code packaged in due to build process.

gcp-unified-mojo-scorer/src/main/java/ai/h2o/mojos/deploy/gcp/unified/config/EnvironmentConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+    downloadFromGcs(env);
+    log.info("Successfully downloaded files from GCS. Starting rest scoring service");
+  }
+
+  private void downloadFromGcs(Map<String, String> env) {
+    downloadFileFromGcs(getFromEnv(env, "MOJO_GCS_PATH"), Paths.get(MOJO_DOWNLOAD_PATH));
+    downloadFileFromGcs(getFromEnv(env, "LICENSE_GCS_PATH"), Paths.get(LICENSE_DOWNLOAD_PATH));
+  }
+
+  private void downloadFileFromGcs(String gcsPath, Path outputPath) {
```
nkpng2k, line 46: so I'm pretty sure these are from my google cloud run code.

nkpng2k (changes requested):
ok:
1. so much copy paste :cry: 
2. if you are going to copy paste, put it to the gcp unified directory instead of common... its not really common code.

nkpng2k:
I also think it might be worthwhile to investigate method overloading... it might be useful in this instance.

gcp-unified-mojo-scorer/src/main/java/ai/h2o/mojos/deploy/gcp/unified/config/EnvironmentConfiguration.java
nkpng2k: you might want to move this to some location in `common` and then just depend on it. Just please make corresponding change in google cloud run too if that is the choice.

nkpng2k (approved):
yay

PR #236, opened 2021-06-30
Mo/gcp vertex

mohamedasni:
- Rename GCP AI Unified Platform to GCP Vertex
- Add option to pass python preprocessing script


gcp-vertex-ai-mojo-scorer/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+The docker image that is built by this project can be pushed to gcr.io and used for scoring
+Driverless AI Mojos in GCP Vertex AI. 
+
+## Building
+
+Make sure you are in the working directory `dai-deployment-templates`. Typing `pwd` in the terminal
+shell should have a similar output to `/my/path/to/dai-deployment-templates`.
+
+You can build the Docker image in two ways. The first only includes the ability to score and the second
```
nkpng2k, line 11: I would maybe be more explicit as to what are the ramifications of this. I.e. different images (and which ones), that python preprocessing requires the java to run python code, etc.

gcp-vertex-ai-mojo-scorer/README.md
```diff
@@ -0,0 +1,112 @@
...
+## Deploying
+
+To deploy the container follow the steps in Google Documentation here to import the model:
+https://cloud.google.com/vertex-ai/docs/general/import-model
+
+There is one requirement for the container. You __MUST__ include the following environment variables:
+* MOJO_GCS_PATH = `gs://path/to/pipeline.mojo`
+* LICENSE_GCS_PATH = `gs://path/to/driverless/ai/license.sig`
+
+If you built the Docker image with the preprocessing script option, you also __MUST__ include the following environment variable:
```
nkpng2k, line 49: **MUST**? I would be careful with this. Even if there is the option to use preprocessing stage, it shouldn't be required.

gcp-vertex-ai-mojo-scorer/README.md
```diff
@@ -0,0 +1,112 @@
...
+* PREPROCESSING_SCRIPT_PATH = `gs://path/to/preprocessing_script.py`
+
+The prediction route will be `/model/score` and the health route will be `/model/id`.
+
+You can then deploy an endpoint once the model has been imported.
+
+## Preprocessing Script
+
+If you are including a Python data preprocessing script, here is an example:
```
nkpng2k, line 58: I would include this in examples dir or something. Gradle will ignore anything that is not in `src` so you are safe to create an `examples` directory and add this there with only a reference to hyperlink

gcp-vertex-ai-mojo-scorer/README.md
```diff
@@ -0,0 +1,112 @@
...
+    [81081.0,-0.873286664580274,0.5383418573374871,2.65055031553525,0.331972130017534,0.0510277957117218,0.6364468979718391,0.6719139650867261,0.18165642237906401,0.0793536531651418,-0.684037044169426,0.8644669396918221,0.8734041370577021,-0.7131054601776229,-0.494781246049671,-1.94685930894166,-0.6014164255583699,-0.0460304794694537,-0.356068770928204,-0.537032123913058,-0.0619970210088985,-0.00642393408573803,0.32916282786391804,-0.33233023038250503,0.22704204810440198,0.546100808417714,-0.40064788122765005,-0.0896039366748501,-0.166155749695546,38.0]
+]
+
+# Replace data with new samples
+request["instances"] = data
+
+# Dump new request to JSON
+with open("/tmp/"+fileName, 'w') as f:
```
nkpng2k, line 81: So I guess we are sticking with disk io for now. No change necessary. But this maybe needs to be made clear in the docs.

gcp-vertex-ai-mojo-scorer/src/main/java/ai/h2o/mojos/deploy/gcp/vertex/ai/config/EnvironmentConfiguration.java
```diff
@@ -41,6 +42,12 @@ public void configureScoringEnvironment() {
   private void downloadFromGcs(Map<String, String> env) {
     downloadFileFromGcs(getFromEnv(env, "MOJO_GCS_PATH"), Paths.get(MOJO_DOWNLOAD_PATH));
     downloadFileFromGcs(getFromEnv(env, "LICENSE_GCS_PATH"), Paths.get(LICENSE_DOWNLOAD_PATH));
+    // Only use pre-processing script if one is provided
+    if (!getFromEnv(env, "PREPROCESSING_SCRIPT_PATH").isEmpty()) {
+      downloadFileFromGcs(
```
nkpng2k, line 47: I feel like this needs error handling?

gcp-vertex-ai-mojo-scorer/src/main/java/ai/h2o/mojos/deploy/gcp/vertex/ai/config/EnvironmentConfiguration.java
```diff
@@ -66,7 +73,7 @@ private Properties parseGcsPath(String gcsPath) {
 
   private String getFromEnv(Map<String, String> env, String envVar) {
     String envValue = env.getOrDefault(envVar, "");
-    if (envValue.isEmpty()) {
+    if (envValue.isEmpty() && !envVar.equals("PREPROCESSING_SCRIPT_PATH")) {
```
nkpng2k, line 76: nit: this seems like some helper method, but there is now a side effect of checking if said arg is `PREPROCESSING_SCRIPT_PATH`, it is better to avoid this entirely I would think.

gcp-vertex-ai-mojo-scorer/src/main/java/ai/h2o/mojos/deploy/gcp/vertex/ai/controller/ModelsApiController.java
```diff
@@ -74,8 +85,40 @@ public static ScoreRequest getRestScoreRequest(
...
+      // Run Python pre processing script and reload request body with updated data
+      ProcessBuilder processBuilder;
+      processBuilder = new ProcessBuilder("python", "/tmp/preprocessing_script.py", fileName);
+      processBuilder.redirectErrorStream(true);
+      log.info("Waiting for script to run");
+      Process process = processBuilder.start();
+      int exitCode = process.waitFor();
+      
+      if (exitCode != 0) {
+        throw new Exception("Preprocessing script failed.");
```
nkpng2k, line 108: is there way to get the actual exception into the code?

gcp-vertex-ai-mojo-scorer/src/main/java/ai/h2o/mojos/deploy/gcp/vertex/ai/controller/ModelsApiController.java
```diff
@@ -74,8 +85,40 @@ public static ScoreRequest getRestScoreRequest(
...
+      }
+      
+      // Read preprocessed data
+      JsonReader reader = new JsonReader(new FileReader("/tmp/" + fileName));
```
nkpng2k, line 112: I'm pretty sure this could be just a byte reader, although I don't know how much it actually changes things. If `JsonReader` is an extra dependency, I'd encourage to not use it. But if its a built in, then I don't care

gcp-vertex-ai-mojo-scorer/src/main/java/ai/h2o/mojos/deploy/gcp/vertex/ai/controller/ModelsApiController.java
```diff
@@ -74,8 +85,40 @@ public static ScoreRequest getRestScoreRequest(
...
+      Gson gson = new GsonBuilder().setPrettyPrinting().create();
+      gcpRequest = gson.fromJson(reader,
+          ai.h2o.mojos.deploy.common.rest.vertex.ai.model.ScoreRequest.class);
+      
+      Files.deleteIfExists(Paths.get("/tmp" + fileName));
```
nkpng2k, line 117: I think this whole thing probably needs some exception handling, or it might be possible to kill the whole jvm

gradle.properties
```diff
@@ -45,6 +45,7 @@ errorproneVersion = 2.3.3
 
 # Docker settings
 dockerRepositoryPrefix = harbor.h2o.ai/opsh2oai/h2oai/
+dockerUsePython = true
```
nkpng2k, line 48: nit: `dockerIncludePython`? wdyt? IMHO `dockerUserPython` means that you will be using python for the build process, as opposed to including it in the actual container image.

gcp-vertex-ai-mojo-scorer/src/main/java/ai/h2o/mojos/deploy/gcp/vertex/ai/config/EnvironmentConfiguration.java
```diff
@@ -41,6 +42,12 @@ public void configureScoringEnvironment() {
   private void downloadFromGcs(Map<String, String> env) {
```
nkpng2k, line 42: nit: maybe this needs to be renamed to something like, `downloadArtifactsFromGcs` that or maybe add a docstring.

nkpng2k (changes requested):
generally looks ok. I left a couple comments, which I think should be considered.

gcp-vertex-ai-mojo-scorer/README.md
```diff
@@ -1,41 +1,68 @@
...
+If the environment variable `PREPROCESSING_SCRIPT_PATH` is not included, then the Docker image will not use
+a preprocessing script. In which case, we highly recommend that you build the image without Python included.
+
+The prediction route will be `/model/score` and the health route will be `/model/id`.
+
 You can then deploy an endpoint once the model has been imported.
 
+## Preprocessing Script
+
+If you are including a Python data preprocessing script, you can find an example in the exampels folder here: [preprocessing_script_example.py](examples/preprocessing_script_example.py)
```
nkpng2k, line 62: typo: `exampels`

gcp-vertex-ai-mojo-scorer/src/main/java/ai/h2o/mojos/deploy/gcp/vertex/ai/controller/ModelsApiController.java
```diff
@@ -74,8 +86,45 @@ public static ScoreRequest getRestScoreRequest(
...
+        Process process = processBuilder.start();
+        process.waitFor();
+        
+        // Read preprocessed data
+        JsonReader reader = new JsonReader(new FileReader("/tmp/" + fileName));
+        Gson gson = new GsonBuilder().setPrettyPrinting().create();
+        gcpRequest = gson.fromJson(reader,
+            ai.h2o.mojos.deploy.common.rest.vertex.ai.model.ScoreRequest.class);
+        
+        Files.deleteIfExists(Paths.get("/tmp" + fileName));
```
nkpng2k, line 116: so I'd probably advise to to a try-with-resources block here: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

probably makes sense to always clean up and delete the file.... for example, what happens if there is an exception during the process. then the file wouldn't always be deleted and you could end up with extra cruft lying around the system taking up space

gcp-vertex-ai-mojo-scorer/src/main/java/ai/h2o/mojos/deploy/gcp/vertex/ai/controller/ModelsApiController.java
```diff
@@ -74,8 +86,45 @@ public static ScoreRequest getRestScoreRequest(
...
+      ObjectMapper mapper = new ObjectMapper();
+      
+      String fileName = UUID.randomUUID().toString();
+      
+      try {
+        mapper.writeValue(new File("/tmp/" + fileName), gcpRequest);
+        
+        // Run Python pre processing script and reload request body with updated data
+        ProcessBuilder processBuilder;
+        processBuilder = new ProcessBuilder("python", "/tmp/preprocessing_script.py", fileName);
```
nkpng2k, line 104: nit: I am not super comfortable with directly calling `python` here... given there is a possibility that future images have maybe different default python or something. I'd say no action necessary here but just for posterity I'm commenting.

nkpng2k (changes requested):
Thanks for the changes, looks a lot better. 2 things: 1 typo (easy), and I'd advise a `try-with-resources` block to avoid accidentally leaving files and open resources lying around.

gcp-vertex-ai-mojo-scorer/src/main/java/ai/h2o/mojos/deploy/gcp/vertex/ai/controller/ModelsApiController.java
nkpng2k: also, there is a possibility of accidentally leaving open readers/writers which can cause problems in garbage cleanup/memory

gcp-vertex-ai-mojo-scorer/src/main/java/ai/h2o/mojos/deploy/gcp/vertex/ai/controller/ModelsApiController.java
```diff
@@ -74,8 +87,47 @@ public static ScoreRequest getRestScoreRequest(
...
+        Gson gson = new GsonBuilder().setPrettyPrinting().create();
+        gcpRequest = gson.fromJson(reader,
+            ai.h2o.mojos.deploy.common.rest.vertex.ai.model.ScoreRequest.class);
+      } catch (JsonSyntaxException e) {
+        log.error("Malformed JSON when reading from file: {}", e.getMessage());
+      } catch (Exception e) {
+        log.error("Unexpected error during data preprocessing step: {}", e.getMessage());
+      } finally {
+        try {
+          Files.deleteIfExists(Paths.get("/tmp" + fileName));
```
nkpng2k, line 122: I mean anything in a finally clause really shouldn't throw an error.... but ok. I think in this context it kind of makes sense.

nkpng2k (approved):
:+1:  Thanks for the changes

PR #242, opened 2021-08-12
[KDB]: upgrade log4j to log4j2

mmalohlava:


achraf-mer (approved):

mmalohlava:
@achraf-mer local run of snyk reports green light - going ahead with merge.

The service seems to be confused by log4j version.

achraf-mer:
> @achraf-mer local run of snyk reports green light - going ahead with merge.
> 
> The service seems to be confused by log4j version.

@mmalohlava thanks, let's also run it by Twistlock in isolation.

mmalohlava:
@achraf-mer probably understand why Snyk is complaining:

image

it seems like service is not able to understand our Gradle configuration mixing - see version is `undefined`.

The second surprising fact is that Snyk service is not complaining about MOJO dependency, however local run complains.

achraf-mer:
One thing to note is that the mojo issue does not have an upgrade path, so unsure if this is what prevents it from showing up on the PR scan.

PR #246, opened 2021-08-21
[DOCS] Add documentation how to enable https for local rest scorer

### LINKED ISSUE #244: Enhance local rest scorer to support HTTPs natively
Support ticket  - https://support.h2o.ai/a/tickets/100029

Customer needs native HTTPS support in rest scorer. We provided a solution fronting scorer with nginix for SSL termination but the customer is seeking a native solution.

nkpng2k:
Resolves #244 

provide some documentation related to https for scorers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ The structure is as follows:
   * `transform`: Shared transformers between mojo and API classes.
 * `{target}-scorer`: One module per deployment target.
 
+Note: Please look to each subproject for additional deployment information.
 
 ## Build
 
--- a/local-rest-scorer/README.md
+++ b/local-rest-scorer/README.md
@@ -25,21 +25,12 @@ java -Dmojo.path={PATH_TO_MOJO_PIPELINE} -jar build/libs/local-rest-scorer-{YOUR
 Springboot application come with built-in support for HTTPS. You can get more information from the
 Springboot [documentation](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#howto.webserver.configure-ssl)
 
-#### Simple Configuration
-
+The servers support using either JKS or PKSC12 type certificates to secure network traffic. 
 If you do not have certificates, you can create a self-signed SSL Certificate
-with java built in tool `keytool` or via `openssl`.
+with java built in tool `keytool` or via `openssl`. See some examples [below](#using-or-creating-self-signed-certificates).
 
-```shell
-# Example: using keytool
-keytool -genkey \
-        -alias selfsigned_localhost_sslserver \
-        -keyalg RSA -keysize 2048 \
-        -validity 700 \
-        -keypass <changeit> \
-        -storepass <changeit> \
-        -keystore ssl-server.jks
-```
+
+#### Simple Configuration
 
 Configuration of SSL requires several parameters to be set for the application:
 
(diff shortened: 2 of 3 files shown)


somanathghalimath:
A couple of observations @nkpng2k, Please let me know if you agree.

Usually when the customers procure SSL certificate package(from a CA or a self-signed utility), they get a certificate, a private key and occasionally a pfx file(required by windows servers). At the bare minimum, they would have a certificate and a private key to start with. 

The approach mentioned here still puts the onus on the customers to convert the certificate/key they procured to jks format that spring-boot requires, so I am not really sure if we can call this a pure native SSL support. 

(Even though you have mentioned a way to generate a self-signed certificate using key-tool, that directly gives the jks file - the most common scenario on field would be customers already having their own set of certificate/key either CA signed or self-sgined for the host they would identify ( @hemenkapadia , please correct me if this is not a right assertion)

I was thinking of an approach where our jar accepts the .crt and .key file from the user and programmatically creates a jks keystore by using the Java Cryptography and Keystore apis, then loads this keystore at the run-time to enable SSL. The approach I was thinking is illustrated here in this linkedin post - https://www.linkedin.com/pulse/creating-java-keystore-programmatically-bill-young, however there is still an open question on the feasibility of loading the jks at the runtime in the embedded tomcat(in spring-boot either by creating a new SSL context or loading the keystore before the tomcat is started) which is something we have to research/test but if that is feasible then I think that would take away any pre-requisite steps needed from the customer and start using the scorer just with the cert/key pair they have.

Please let me know your thoughts on this.

nkpng2k:
@somanathghalimath considering that this is a java application, I don't feel particularly bad about asking the user to provide java keystore. @hemenkapadia and I decided on this route based on internal discussion on slack. 

My thought is as follows:
1. This is Native support. It is a built in support of https with provided certificates, as provided by the server. 
2. If the user is requesting a set of certificates, what prevents them from requesting a JKS cert instead of pem certs? Ultimately they are still created/provided by the IT team (I would expect)
3. Sure we can add support for pem files. It is not particularly difficult, but if the problem statement is "provide native support for ssl" then that has been achieved imho.

hemenkapadia:
My understanding (based on interaction with the customer)  of the term "native" is that the REST Scorer application should itself provide HTTPs connection capability when specific command-line arguments are passed. The "native" meant that he did not want to front the REST Scorer with a Nginix/Apache/HAProxy etc.. for just SSL termination.

I do not have a hard inclination if the REST Scorer command line should accept cert/key or the JKS, but on the field, cert/key is more prevalent. If we are going to implement the scorer to accept JKS only on the command line, that is ok. 

But then we need to provide instructions on how to convert the existing cert/key to jks and use it with our tool, something like discussed at https://stackoverflow.com/questions/906402/how-to-import-an-existing-x-509-certificate-and-private-key-in-java-keystore-to

I don't think any one uses self signed certs on the field for actual work.

somanathghalimath:
I agree @nkpng2k. I think that makes sense. 
I would recommend to incorporate steps for cert/key conversion to jks, something like below

- Convert the SSL certificate and the Key into the pk12 format using the below command

`openssl pkcs12 -export -in ​server-certificate.pem -inkey ​server-key.pem -name scorer -out scorer.p12
`
This would ask for a password, do make a note of it and provide it in the next command

- Create a Java key store using the below command, pass an appropriate password in the command below (shown in angular brackets) which will be passed as a start up parameter to the scorer

`keytool -importkeystore -deststorepass <password for jks store> -destkeystore scorer.jks -srckeystore scorer.p12 -srcstoretype PKCS12
`
When prompted for the source key store password, please enter the password from step 1

nkpng2k:
Ok this is fair. Will update. Although, I'd argue that this is fundamental stuff that anyone can google/look up.

somanathghalimath:
You are right @nkpng2k, the official spring-boot documentation briefly mentions jks as the requirement for enablement of SSL and that was the direction I followed to convert cert/key pair to JKS without realizing the intermediate pk12 trust store works as well :)

mmalohlava:
@nkpng2k thank you! i have a nit pick comment - it would be good to have test scenario for this documentation.

nkpng2k:
:smirk: since I already merged, I will open new pr with this later.

PR #250, opened 2021-08-30
Changes to support shapley values for original features

Rajimut:


common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -46,8 +43,8 @@
   // note the mojo pipeline need to be reloaded here as we have a constrain from java mojo
   // both SHAP values and predictions cannot be provided with the same pipeline
   // Link: https://github.com/h2oai/mojo2/blob/7a1ab76b09f056334842a5b442ff89859aabf518/doc/shap.md
-  private static final MojoPipeline pipelineShapley = loadMojoPipelineFromFile();
-
+  private static final MojoPipeline pipelineTransformedShapley = loadMojoPipelineFromFile();
```
nkpng2k, line 46: :scream: we load the mojo 3 times now? We shouldn't really be loading the mojo more than once. Like what happens if the user provides a 3GB mojo -- then each load == 36+ GB in memory * 3 = 108 GB just to have scorer for one mojo. Apologies, I should have caught this in the original pr, imho this really shouldn't happen. cc @mmalohlava wdyt? maybe I'm misunderstanding how mojo is loaded?

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -230,13 +235,18 @@ private static MojoFrame doScore(MojoFrame requestFrame) {
...
+  private static MojoFrame doShapleyContrib(MojoFrame requestFrame, boolean isOriginal) {
     log.debug(
             "Input has {} rows, {} columns: {}",
             requestFrame.getNrows(),
             requestFrame.getNcols(),
             Arrays.toString(requestFrame.getColumnNames()));
-    MojoFrame shapleyResponseFrame = pipelineShapley.transform(requestFrame);
+    MojoFrame shapleyResponseFrame;
+    if (isOriginal) {
+      shapleyResponseFrame = pipelineOriginalShapley.transform(requestFrame);
```
nkpng2k, line 246: regarding my earlier statement about loading mojos. Given that we are scoring in sequence: 
1. score predictions
2. score shapley if requested
3. etc.
Why don't we just do:
```
pipeline.setShapPredContrib(true);
MojoFrame shapleyResponseFrame = pipeline.transform(requestFrame);
pipeline.setShapPredContrib(false);
```

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -136,19 +140,20 @@ public ContributionResponse computeContribution(ContributionRequest request) {
     ShapleyType requestedShapleyType = shapleyType(request.getRequestShapleyValueType());
     switch (requestedShapleyType) {
       case TRANSFORMED:
-        MojoFrame requestFrame = contributionRequestConverter
-                .apply(request, pipelineShapley.getInputFrameBuilder());
-        return contribution(requestFrame);
+        MojoFrame requestFrame1 = contributionRequestConverter
```
nkpng2k, line 143: nit: how is `requestFrame1` different from `requestFrame2`? seems like the it is a result which is returned without being carried past the switch. I think probably it is better to have just `requestFrame`

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -96,10 +94,10 @@ public ScoreResponse score(ScoreRequest request) {
       ShapleyType requestedShapleyType = shapleyType(request.getRequestShapleyValueType());
       switch (requestedShapleyType) {
         case TRANSFORMED:
```
nkpng2k, line 96: qq - was there an option for `BOTH` ever or was it decided that giving both results was too much?

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
Rajimut: yes, because the mojo pipeline/runtime still doesn't support this.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
Rajimut: Currently we cannot change the boolean on the runtime, unfortunately the pipeline doesn't support this
https://github.com/h2oai/mojo2/blob/7a1ab76b09f056334842a5b442ff89859aabf518/doc/shap.md
> either SHAP values or predictions can be provided, not both at the same time (with the same pipeline)

nkpng2k (changes requested):
One high level note, we should maybe hold off on releasing this until 1.10.0 is actually released.

General approach looks good to me. But very concerned about the fact that we load the mojo into memory multiple times.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
nkpng2k: yea nothing says we cannot set the boolean back to `false` afterwards. @Rajimut

Rajimut: I tested it, reversing doesn't seem to work

nkpng2k: as far as I can tell, its just a setter. See here:
https://github.com/h2oai/mojo2/blob/8c7a253b1a6c886eca2be51b3e6eaac19a69e158/java/mojo2-runtime-impl/src/main/java/ai/h2o/mojos/runtime/MojoPipelineProtoImpl.java#L198-L205
and
https://github.com/h2oai/mojo2/blob/8c7a253b1a6c886eca2be51b3e6eaac19a69e158/java/mojo2-runtime-impl/src/main/java/ai/h2o/mojos/runtime/MojoPipelineProtoImpl.java#L110-L135

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
nkpng2k: also, if this really is the case, then we should be much more careful about this. MLops does not yet have resource control (i.e. being able to define the amount of cpus and memory assigned to a deployment) this could very easily break.

nkpng2k: cc @orendain what could be the impact of trying to do this in mlops? I am personally scared myself. How do you think kubernetes would respond in the instance that we try to deploy an extraordinarily large mojo?

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
nkpng2k: 😭 yea I tried it my self. Raises exception

orendain (approved):
LGTM.

I agree with @nkpng2k that it's concerning that underlying MOJO2 loads the MOJO multiple times.  Until MOJO2 is modified, we'll want to use this next version of REST scorer judiciously.

> One high level note, we should maybe hold off on releasing this until 1.10.0 is actually released.

DAI 1.10.0?  Is DAI blocking this repo from a new version?

nkpng2k:
> LGTM.
> 
> I agree with both @nkpng2k and @Rajimut that it's concerning that underlying MOJO2 loads the MOJO multiple times. Until MOJO2 is modified, we'll want to use this next version of REST scorer judiciously.
> 
> > One high level note, we should maybe hold off on releasing this until 1.10.0 is actually released.
> 
> DAI 1.10.0? Is DAI blocking this repo from a new version?
> 
> Edit: Note to self: v`1.1.3` is the last release of this scorer w/o Shapley values.

@orendain @mmalohlava @Rajimut wdyt if we have a flag to enable / disable this? If its some env var it can be set in byom as a new runtime type no? And this way I believe the user would be able to choose at deploy time 
```
    {
      name = "local-rest-scorer-shapley"
      display-name = "H2O.ai Local Rest Scorer Shapley"
      model-type = "dai/mojo-pipeline"
      image = "local-rest-scorer-image"
      writable-volume = false
      env-vars = [
        { name = "H2OAI_SCORER_ENABLE_SHAPLEY", value = "true" }
      ]
    },
```

Rajimut:
@mmalohlava - your thoughts on the above discussion please..

nkpng2k:
@orendain I believe that 1.10 is the only version that will have mojo shapley support for original columns. 1.9.3+ will have support for transformed columns... afaik. Therefore, I suggest either we merge, but cut the release 1.1.4 from different hash (for support before 1.10 is released), else we don't merge until 1.10 is released

nkpng2k:
cc @mwysokin https://github.com/h2oai/dai-deployment-templates/pull/250#issuecomment-909354106
 if you don't mind chiming in as well?

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
Rajimut: I don't think mojo2 supports both the transformed and original at the same time yet

Rajimut: ref https://github.com/h2oai/mojo2/blob/643128c00f511ca7c9c38abdf4598f3edc373541/java/mojo2-cli/src/main/java/ai/h2o/mojos/ExecuteMojo.java#L142

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -96,10 +94,10 @@ public ScoreResponse score(ScoreRequest request) {
...
-          response.setFeatureShapleyContributions(transformedFeatureContribution(request));
+          response.setFeatureShapleyContributions(featureContribution(request, false));
           break;
         case ORIGINAL:
-          log.info(UNIMPLEMENTED_MESSAGE);
+          response.setFeatureShapleyContributions(featureContribution(request, true));
```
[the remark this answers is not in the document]
Rajimut: I think we need to know what other DAI models doesn't support Shap values. Is there anything apart from regression/binomial/multinomial models? Also how do we handle older DAI models that doesn't have the original matrix. Can we throw error to the score endpoint if requested with Shap? @mmalohlava might need to help us on this.

orendain:
> @orendain I believe that 1.10 is the only version that will have mojo shapley support for original columns. 1.9.3+ will have support for transformed columns... afaik. Therefore, I suggest either we merge, but cut the release 1.1.4 from different hash (for support before 1.10 is released), else we don't merge until 1.10 is released

Ah, so in other words you want to make sure we have a release of the scorer in its current state before this PR is merged.

If so, then IMHO that means a new version of the scorer should be cut now (or at least from some specific tag/hash before this branch), instead of queuing this PR behind a DAI release.

May even be beneficial to have a scorer w/ only transformed Shap values supported be officially cut instead of jumping immediately to a scorer w/ full support.

orendain:
> @orendain @mmalohlava @Rajimut wdyt if we have a flag to enable / disable shapley? If its some env var it can be set in byom as a new runtime type no? And this way I believe the user would be able to choose at deploy time
> 

Ooooh, great idea!  I'd support that.  Even w/o thinking about it's use in other stacks, having the ability for users to disable the 3x MOJO loads in standalone mode of this scorer seems incredibly beneficial.

Perhaps even having Shap values disabled by default, until MOJO2 library implements a lower memory overhead approach, would be ideal.  That would save users of this scorer from having to be explicitly warned about the sudden increase of resource use for moving up a single minor version `1.1.3` -> `1.1.4`.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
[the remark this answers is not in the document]
orendain: > Is there anything apart from regression/binomial/multinomial models?

For DAI MOJOs, for sure.

---

IMHO, having to take stock of the full set of allowed/disallowed set of DAI and H2O-3 models, instead of failing-fast when a dependent library fails, may suggest that we're doing something in a particularly difficult-to-maintain way.

There are numerous ways for scoring requests to fail (for both DAI and H2O-3 MOJOs) once the scorer hands the task off to the underlying MOJO2 lib.  Even in the case of Shap scoring, I imagine there are cases wherein users might need to be made aware that something went wrong, since the issue wasn't a lack of support, but rather a legitimate error.

nkpng2k:
> Ah, so in other words you want to make sure we have a release of the scorer in its current state before this PR is merged.

well not only this, but also, we would immediately break `master` if we merge this for all users until we get a new mojo runtime dependency

nkpng2k:
@Rajimut I am amending my request. Given that it is not supported, and we NEED to have multiple mojo loads:
we need an opt in option. It is too dangerous to ALWAYS load multiple mojos into memory. So let us have a flag like `-Dh2oai.scorer.enable.shapley` and an env var of same name `H2OAI_SCORER_ENABLE_SHAPLEY` to which we only load multiple mojos if `true`

Rajimut:
@nkpng2k - We can do the env variable idea 
Also I think the latest 2.7 has the support for original shap values here 
https://search.maven.org/search?q=a:mojo2-runtime-impl

mwysokin:
@nkpng2k @mmalohlava DAI 1.10 was released. Do we need to make any additional changes to this PR or is it ready to be merged?

nkpng2k:
So looking back I think there is one outstanding things imho:
1. if we do this with current env vars, we'd either load the mojo 3 times or 1 time, so memory consumption could be through the roof. AFAIK there has been no progress on mojo2 side to resolve this. I'd suggest we split the env var into an enum so that we can optionally load the mojo 1/2/3 times depending on specifications. Ex. `ENABLE_SHAPLEY=ORIGINAL,TRANSFORMED,ALL`

According to docs, my initial blocker that this feature wasn't released is no longer true: https://docs.h2o.ai/driverless-ai/1-10-lts/docs/userguide/release_notes.html#version-1-10-0-september-29-2021

So @mwysokin I think we can optionally merge... WDYT?

mwysokin (approved):
SGTM @nkpng2k 🖖 

If we're going to waste memory at least let's give people some options. 

image

nkpng2k (approved):

nkpng2k:
merging based on offline discussion

PR #252, opened 2021-09-02
added a boolean to enable shapley

Rajimut:


common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -92,6 +99,16 @@ public ScoreResponse score(ScoreRequest request) {
     MojoFrame responseFrame = doScore(requestFrame);
     ScoreResponse response = scoreResponseConverter.apply(responseFrame, request);
     response.id(pipeline.getUuid());
+
+    if (request.getRequestShapleyValueType() == null) {
+      return response;
+    }
```
orendain, line 105: If someone has Shap values disabled, and explicitly specifies `NONE` for Shap values, I'd imagine some would argue that the request should succeed.

The additions in this branch would fail those requests.

Rajimut: Fixed this, thanks @orendain

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -42,12 +43,10 @@
   private static final String MOJO_PIPELINE_PATH_PROPERTY = "mojo.path";
   private static final String MOJO_PIPELINE_PATH = System.getProperty(MOJO_PIPELINE_PATH_PROPERTY);
   private static final MojoPipeline pipeline = loadMojoPipelineFromFile();
-
-  // note the mojo pipeline need to be reloaded here as we have a constrain from java mojo
-  // both SHAP values and predictions cannot be provided with the same pipeline
-  // Link: https://github.com/h2oai/mojo2/blob/7a1ab76b09f056334842a5b442ff89859aabf518/doc/shap.md
-  private static final MojoPipeline pipelineShapley = loadMojoPipelineFromFile();
-
+  private static final String H2OAI_SCORER_ENABLE_SHAPLEY_PROPERTY = "h2oai.scorer.enable.shapley";
```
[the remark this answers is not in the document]
Rajimut: renamed it to SHAPLEY_ENABLE_PROPERTY and `shapley.enable`

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -77,7 +80,15 @@ public MojoScorer(
...
     this.csvConverter = csvConverter;
-    pipelineShapley.setShapPredictContrib(true);
+
+    if (ENABLE_SHAPLEY_CONTRIBUTION) {
+      // note the mojo pipeline need to be reloaded here as we have a constrain from java mojo
+      // both SHAP values and predictions cannot be provided with the same pipeline
+      // Link: https://github.com/h2oai/mojo2/blob/7a1ab76b09f056334842a5b442ff89859aabf518/doc/shap.md
+
+      pipelineTransformedShapley = loadMojoPipelineFromFile();
+      pipelineTransformedShapley.setShapPredictContrib(true);
```
orendain, line 90: Is this constructor called more than once during the life of a single scorer run?  Just checking to make sure that the MOJO isn't loaded more than once.

orendain (approved):
LGTM!  Is it feasible to add a test or two?

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
Rajimut: Its called only once from the local rest scorer from the ScorerConfiguration.. But it is also called once from other scorers like gcp's vertex ai and sagemaker.

Rajimut:
@mmalohlava / @mwysokin - This PR is ready to be merged. Can I have the final thoughts?

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -42,12 +47,10 @@
   private static final String MOJO_PIPELINE_PATH_PROPERTY = "mojo.path";
```
mmalohlava, line 47: nit pick: i prefer to prefix properties with a context prefix (e.g., for mojos we are using `ai.h2o.mojo`)

Rajimut: @mmalohlava - this is an existing property that we already set up in runtime, changing this might affect some customers who are already consuming it

mmalohlava (approved):

PR #265, opened 2021-11-08
Update springfox-swagger to use version 3.0.0

achraf-mer:
Changes:
- Upgrade swagger library as per https://github.com/springfox/springfox#spring-boot-applications-1
- Modify doc url from `../swagger-ui.html` to `../swagger-ui/index.html`
- Add guava explicit dependency 

Using version 3.0.0 is needed so can use the newer swagger-ui in [here](https://github.com/springfox/springfox/blob/3.0.0/springfox-swagger-ui/build.gradle#L24), instead of this [old](https://github.com/springfox/springfox/blob/2.9.2/springfox-swagger-ui/build.gradle#L32) version.


mmalohlava (approved):
LGMT

mmalohlava:
@orendain any concerns as you are end user?

PR #270, opened 2021-12-02
Adds ability to optionally load mojos for each Shapley type

nkpng2k:
Adds new property `shapley.types.enabled` which can be one of `[ALL, NONE, ORIGINAL, TRANSFORMED]`

Behavior is as follows:
1. if `ALL` - loads mojo for scoring, original shapley values, and transformed shapley values
2. `ORIGINAL` - loads mojo for scoring and original shapley values
3. `TRANSFORMED` - loads mojo for scoring and transformed shapley values
4. `NONE` - Shapley scoring not enabled
5. set property `shapley.enable` - loads mojo for scoring, original shapley values, and transformed shapley values


common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -47,6 +48,9 @@
   private static final String SHAPLEY_ENABLE_PROPERTY = "shapley.enable";
   private static final boolean ENABLE_SHAPLEY_CONTRIBUTION =
           Boolean.getBoolean(SHAPLEY_ENABLE_PROPERTY);
+  private static final String SHAPLEY_ENABLED_TYPES_PROPERTY = "shapley.types.enabled";
```
orendain, line 51: Awesome work keeping current env var around for backward-compatibility 👍🏾 👍🏾 !

In case you're looking for a trick to not having to keep `ENABLE_SHAPLEY_CONTRIBUTION` around after initial check, some type of logic that folds this into `ENABLED_SHAPLEY_TYPES` could be:

(psuedocode, using ternary only as example)
```
SHAPLEY_ENABLED_TYPES_PROPERTY  = ENABLE_SHAPLEY_CONTRIBUTION ? "ALL" : howeverShaplyEnabledTypesIsUsuallySet()
```

And only `SHAPLEY_ENABLED_TYPES_PROPERTY` afterward.

local-rest-scorer/README.md
```diff
@@ -20,7 +20,22 @@ java -Dmojo.path={PATH_TO_MOJO_PIPELINE} -jar build/libs/local-rest-scorer-{YOUR
 
 To run the local scorer with shapley contribution enabled
 ```bash
+# Equivalent to using below and setting `-Dshapley.types.enabled=ALL`
 java -Dmojo.path={PATH_TO_MOJO_PIPELINE} -Dshapley.enable=true -jar build/libs/local-rest-scorer-{YOUR_CURRENT_VERSION}-boot.jar
+
+# SHAPLEY_TYPE can be one of [ALL, TRANSFORMED, ORIGINAL]
+java -Dmojo.path={PATH_TO_MOJO_PIPELINE} -Dshapley.types.enabled={SHAPLEY_TYPE} -jar build/libs/local-rest-scorer-{YOUR_CURRENT_VERSION}-boot.jar
```
orendain, line 27: Ah, I see now what you mean re: not parsing env vars.  Great point.

https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#getenv(java.lang.String)
In case you want to check for env vars.

It's [good practice](https://12factor.net/config).  Plus we can't reliably pass in arbitrary system properties from MLOps, but we can for envVars since it's standard method of configuration and not specific to Java.

orendain (commented):
Awesome, just a couple of points that might be worth thinking about.

QQ:  If a user has only original enabled, but sends in a request for transform, what do you define should be the scorer's behavior?  ATM I think requests continue through.

common/swagger/swagger.yaml
```diff
@@ -343,7 +343,7 @@ definitions:
           $ref: '#/definitions/DataField'
   ShapleyType:
     type: string
-    enum: [ ORIGINAL, TRANSFORMED, NONE ]
+    enum: [ ORIGINAL, TRANSFORMED, NONE, ALL ]
```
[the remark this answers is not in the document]
nkpng2k: hmmm yea this is a good point. Probably makes sense to maybe use internally recognized enum instead.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -47,6 +48,9 @@
...
+  private static final ShapleyType ENABLED_SHAPLEY_TYPES =
+          ShapleyType.fromValue(System.getProperty(SHAPLEY_ENABLED_TYPES_PROPERTY, "NONE"));
```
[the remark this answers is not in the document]
nkpng2k: ya agree with your latter statement. I think thats what I'm going to do.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -299,6 +293,67 @@ public Model getModelInfo() {
     return modelInfoConverter.apply(pipeline);
   }
 
+  private boolean isShapleyEnabled() {
```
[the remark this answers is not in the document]
nkpng2k: +1 for setting in init

local-rest-scorer/README.md
nkpng2k: ya... I realized as much as I was reading the previous implementation. Imho would be nice to have as another refactor maybe... I think I will separate reading from env var to a separate pr, simply because its only partially related to this.

nkpng2k:
> Awesome, just a couple of points that might be worth thinking about.
> 
> QQ: If a user has only original enabled, but sends in a request for transform, what do you define should be the scorer's behavior? ATM I think requests continue through.

Ya so at this specific moment it would pass it through. I suppose some logic is worth adding....

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ShapleyLoadOptions.java
```diff
@@ -0,0 +1,69 @@
...
+  public boolean isEnabled() {
+    ShapleyLoadOptions options = ShapleyLoadOptions.fromValue(value);
+    switch (options) {
+      case ALL:
+      case ORIGINAL:
+      case TRANSFORMED:
+        return true;
+      case NONE:
+      default:
+        return false;
```
orendain, line 40: I can see benefit of listing each case, just adding here to lend a fresh pair of eyes.  May be reducible to `return options != NONE`;

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ShapleyLoadOptions.java
```diff
@@ -0,0 +1,69 @@
+package ai.h2o.mojos.deploy.common.transform;
+
+/**
+ * Enum defining options for loading the mojo to enable Shapley predictions.
+ */
+public enum ShapleyLoadOptions {
```
orendain, line 6: nit: Traditionally, enum is named after the representation of the single type and not the collective.

orendain (approved):
LGTM, just some optimizations for consideration.

It may be wise to add a few tests.  For reference, a previous shapley PR added a handful to: https://github.com/h2oai/dai-deployment-templates/blob/master/common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/MojoScorerTest.java

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ShapleyLoadOptions.java
```diff
@@ -0,0 +1,69 @@
...
+  ALL("ALL"),
+  NONE("NONE"),
+  ORIGINAL("ORIGINAL"),
```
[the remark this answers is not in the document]
nkpng2k: Well I partially opted for extended enum because it means that we can have one centralized location for shapley as opposed to each scorer implementing its own version of this. In theory our lamda/sagemaker/other deployments in this repo could implement shapley scoring as well.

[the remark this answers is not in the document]
nkpng2k: but I like a lot of your comments. Going to shoot for a quick refactor.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -299,6 +304,41 @@ public Model getModelInfo() {
...
+  /**
+   * Method to load mojo pipelines for shapley scoring based on configuration
+   *
+   * <p>Order of operations to preserve backwards compatibility:
+   * 1. if property or env var shapley.types.enabled is set, load pipelines based on that
+   * 2. if shapley.enabled is true load all pipelines
+   *
+   */
+  private void loadMojoPipelinesForShapley() {
+    if (!ShapleyLoadOptions.NONE.equals(ENABLED_SHAPLEY_TYPES)) {
```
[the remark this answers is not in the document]
nkpng2k: as far as I can tell, the `.equals` here is an enum comparison. I don't think it has any underlying `toString()` ... at least I checked and the base `Enum` class has an equality method that is just returning `==`

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -299,6 +309,42 @@ public Model getModelInfo() {
...
+  /**
+   * Method to load mojo pipelines for shapley scoring based on configuration
+   *
+   * <p>Order of operations to preserve backwards compatibility:
+   * 1. if property or env var shapley.types.enabled is set, load pipelines based on that
+   * 2. if shapley.enabled is true load all pipelines
+   *
+   */
+  private void loadMojoPipelinesForShapley() {
+    if (ShapleyLoadOption.NONE.equals(ENABLED_SHAPLEY_TYPES)) {
```
orendain, line 321: From `Effective Java` by Joshua Bloch.

>  Item 1: Consider static factory methods instead of constructors
[...] it allows an immutable class to make the guarantee that no two equal instances exist: a.equals(b) if and only if a==b. If a class makes this guarantee, then its clients can use the == operator instead of the equals(Object) method, which may result in improved performance. Enum types provide this guarantee.

Improved performance, null safe, and compile type checking.

`enum1 == "Nick"` -- compiler complains

vs

`enum1.equals("Nick")` -- uh oh, compiles

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ShapleyLoadOption.java
```diff
@@ -0,0 +1,41 @@
...
+  /**
+   * Checks whether requested type of Shapley value scoring is permitted.
+   * @param requested {@link String}
+   * @return {@link Boolean}
+   */
+  public static boolean requestedTypeEnabled(ShapleyLoadOption option, String requested) {
+    if (option.equals(ALL)) {
+      return true;
+    }
+    return option.name().equals(requested);
```
orendain, line 39: Not required, just adding because it's sometimes fun to see how concise a return statement can be:

`return option == ALL || option.name().equals(requested);`

But arguably easier to read as you have it.  Just sharing.

common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/MojoScorerTest.java
```diff
@@ -136,35 +134,326 @@ void verifyScoreRequestWithOriginalShapley_fails() {
...
+    ScoreResponse dummyResponse = generateDummyResponse();
+    given(scoreResponseConverter.apply(any(), any()))
+        .willReturn(dummyResponse);
+
+    MojoScorer scorer = dummyScorer();
+
+    // When
+    scorer.score(request);
+
+    // Then all ok
```
orendain, line 195: Do we know all is okay because no exception was thrown by scorer.score()?  If so, then asserting with something like:
```
// When & Then
assertDoesNotThrow(...)
```

common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/MojoScorerTest.java
```diff
@@ -136,35 +134,326 @@ void verifyScoreRequestWithOriginalShapley_fails() {
...
```
orendain, line 290: In the specific-shapley tests, do we need to test that the scorer is indeed only recognizing/returning what we want?

Due to time constraint, I confess I am not looking very closely at how the test class props are set.  I'm not sure if your props/mocks allow making these verifications.

orendain (approved):
Still LGTM, with a couple of Java convention tips:

Now that a few fields are no longer constant, check out comment about changing `static` identifier + camelCase.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -86,6 +82,14 @@ public MojoScorer(
     this.modelInfoConverter = modelInfoConverter;
     this.csvConverter = csvConverter;
 
+    ENABLED_SHAPLEY_TYPES =
+        Boolean.getBoolean(SHAPLEY_ENABLE_PROPERTY)
+            ? ShapleyLoadOption.ALL
+            : ShapleyLoadOption.valueOf(
+            System.getProperty(SHAPLEY_ENABLED_TYPES_PROPERTY, "NONE"));
+    SHAPLEY_ENABLED =
```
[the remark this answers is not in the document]
nkpng2k: yea... I think maybe I cheated on this? basically in the tests, the values above were never changed based on the set system properties.... like at least in the way we reinstantiate the scorer per test.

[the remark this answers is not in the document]
nkpng2k: like I'd happily revert this ... i just didn't find a better option.

common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/MojoScorerTest.java
nkpng2k: ... well at the moment yes.

nkpng2k: tbh I noticed we kind of consume the exception if anything DOES happen... which is kind of not ok.

common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/MojoScorerTest.java
nkpng2k: I think its fair for now... TBH I'm mostly stealing from tests that already preexisted and just extending for additional cases. Eventually it would probably make sense to have more indepth tests but we also consume the exception if there is one. So I'm inclined to say this is good enough for now.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ShapleyLoadOption.java
nkpng2k: hehe so when I have to parse `||` in my head it immediately becomes harder to read :rofl:

nkpng2k: so yea I'm going to stick with readability.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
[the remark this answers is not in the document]
nkpng2k: I think I simply removed the wrong part. They are constants `final` is appropriate, but not `static`

common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/MojoScorerTest.java
nkpng2k: We could mock the response out of the method to actually return some expected value I suppose.

nkpng2k: So afaik we cannot mock private methods... and the public exposed stuffs doesn't really provide the ability for us to check anything else except that it doesn't barf... So really unless I dive deeper under the hood this is probably as good as it gets.

PR #271, opened 2021-12-07
Adds support for configuring shapley value scoring via environment variable

nkpng2k:
as per comment here: https://github.com/h2oai/dai-deployment-templates/pull/270#discussion_r762343537

we need to support setting environment variables to configure shapley scoring for mlops purposes. 

TBH, I'd like to consider something like picocli to configure this moving forwards, but for now, this provides decent functionality as well as the ability for all our scorers here to implement shapley if necessary.

### DIFF
--- a/common/transform/build.gradle
+++ b/common/transform/build.gradle
@@ -14,7 +14,7 @@ dependencies {
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api'
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-params'
     testRuntimeOnly group: 'org.junit.jupiter', name: 'junit-jupiter-engine'
-    testImplementation group: 'uk.org.webcompere', name: 'system-stubs-jupiter', version: '1.2.0'
+    testImplementation group: 'uk.org.webcompere', name: 'system-stubs-jupiter'
 }
 
 test {
--- a/common/transform/build.gradle
+++ b/common/transform/build.gradle
@@ -14,6 +14,7 @@ dependencies {
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api'
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-params'
     testRuntimeOnly group: 'org.junit.jupiter', name: 'junit-jupiter-engine'
+    testImplementation group: 'uk.org.webcompere', name: 'system-stubs-jupiter', version: '1.2.0'
 }
 
 test {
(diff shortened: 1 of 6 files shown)


common/transform/build.gradle
```diff
@@ -14,6 +14,7 @@ dependencies {
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api'
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-params'
     testRuntimeOnly group: 'org.junit.jupiter', name: 'junit-jupiter-engine'
+    testImplementation group: 'uk.org.webcompere', name: 'system-stubs-jupiter', version: '1.2.0'
```
orendain, line 17: Consider dropping the version here, and instead adding version to https://github.com/h2oai/dai-deployment-templates/blob/master/gradle.properties and https://github.com/h2oai/dai-deployment-templates/blob/master/gradle/mixins/dependencies.gradle.

That way, dependency versions are centralized and shared across projects.

common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/ShapleyLoadOptionTest.java
```diff
@@ -0,0 +1,143 @@
...
+    // Given
+    System.setProperty("shapley.types.enabled", "TRANSFORMED");
+
+    // When
+    ShapleyLoadOption option = ShapleyLoadOption.fromEnvironment();
+
+    // Then
+    assertEquals(ShapleyLoadOption.ORIGINAL, option);
+  }
+}
```
orendain, line 143: nit: newline

common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/ShapleyLoadOptionTest.java
```diff
@@ -0,0 +1,143 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import uk.org.webcompere.systemstubs.environment.EnvironmentVariables;
+import uk.org.webcompere.systemstubs.jupiter.SystemStub;
+import uk.org.webcompere.systemstubs.jupiter.SystemStubsExtension;
+
+@ExtendWith(SystemStubsExtension.class)
+public class ShapleyLoadOptionTest {
```
orendain, line 15: nit: No need to make test class public.  Can leave out that modifier and just have `class ...`

orendain (approved):
LGTM - awesome tests.  Thanks for introducing me to the env var testing library.

PR #276, opened 2021-12-21
Push scorer images to gcr

achraf-mer:
image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -43,9 +43,9 @@ pipeline {
             description: 'Whether to also push distribution ZIP archive to S3.',
         )
         booleanParam(
-                name: 'PUSH_TO_GOOGLE_CLOUD',
+                name: 'PUSH_TO_VORVAN',
                 defaultValue: false,
-                description: 'Whether to also push Docker images to Google Cloud Registry.',
+                description: 'Whether to also push Docker images to h2o.ai maintained gcr.io repo Vorvan.',
         )
     }
 
@@ -222,7 +222,7 @@ pipeline {
         stage('7. Push Docker Images To GoogleCloud') {
             when {
                 expression {
-                    return isReleaseBranch() || params.PUSH_TO_GOOGLE_CLOUD
+                    return isReleaseBranch() || params.PUSH_TO_VORVAN
                 }
             }
             agent {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,6 +9,7 @@ NODE_LABEL = 'docker'
 DOCKERHUB_CREDS = 'dockerhub'
 HARBOR_URL = "http://harbor.h2o.ai/"
 HARBOR_CREDS = 'harbor.h2o.ai'
+VORVAN_CRED = 'MM_GCR_VORVAN_CREDENTIALS'
 
 def versionText = null
 def utilsLib = new Utils()
@@ -40,6 +41,11 @@ pipeline {
             name: 'PUSH_DISTRIBUTION_ZIP',
             defaultValue: false,
             description: 'Whether to also push distribution ZIP archive to S3.',
+        )
+        booleanParam(
+                name: 'PUSH_TO_GOOGLE_CLOUD',
+                defaultValue: false,
+                description: 'Whether to also push Docker images to Google Cloud Registry.',
         )
     }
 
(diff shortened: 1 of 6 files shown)


Jenkinsfile
```diff
@@ -212,6 +218,42 @@ pipeline {
...
+            }
+            steps {
+                timeout(time: 30, unit: 'MINUTES') {
+                    script {
+                        def gitCommitHash = env.GIT_COMMIT
+                        def imageTags = "${versionText},${gitCommitHash}"
+                        withDockerCredentials(DOCKERHUB_CREDS, "FROM_") {
+                            withGCRCredentials(VORVAN_CRED) {
+                                def gcrCreds = readFile("${GCR_JSON_KEY}")
+                                withEnv(['TO_DOCKER_USERNAME=_json_key', "TO_DOCKER_PASSWORD=${gcrCreds}"]) {
```
achraf-mer, line 244: tested this a couple of times, if inlined with gradle command the file content will show up on jenkins console logs (not ideal)

Jenkinsfile
```diff
@@ -41,6 +42,11 @@ pipeline {
             defaultValue: false,
             description: 'Whether to also push distribution ZIP archive to S3.',
         )
+        booleanParam(
+                name: 'PUSH_TO_GOOGLE_CLOUD',
```
nkpng2k, line 46: nit: for naming consistency, probably `push_to_vorvan` makes more sense... although since this is public repo, I'd also say putting to the description that its going to `h2o.ai maintained gcr.io repo vorvan` would also be ok.

Jenkinsfile
nkpng2k: maybe something like this? https://github.com/h2oai/mlops-byom-images/blob/main/Jenkinsfile#L176-L193

nkpng2k: and have 2 separate stages?

nkpng2k (approved):
Generally lgtm, small thoughts only

Jenkinsfile
mwysokin: @nkpng2k we do unfortunately, because docker fetches the base img usually form dockerhub or just checks its signature, we actually had issues with pushing to gcr when dockerhub was not authenticated...

Jenkinsfile
```diff
@@ -212,6 +218,42 @@ pipeline {
...
+            }
+            steps {
+                timeout(time: 30, unit: 'MINUTES') {
+                    script {
+                        def gitCommitHash = env.GIT_COMMIT
+                        def imageTags = "${versionText},${gitCommitHash}"
+                        withDockerCredentials(DOCKERHUB_CREDS, "FROM_") {
+                            withGCRCredentials(VORVAN_CRED) {
+                                def gcrCreds = readFile("${GCR_JSON_KEY}")
+                                withEnv(['TO_DOCKER_USERNAME=_json_key', "TO_DOCKER_PASSWORD=${gcrCreds}"]) {
```
mwysokin, line 244: just in case you'd like to do it consistently as jib flags, not to have some things taken from env and some from flags, a version with only flags used can be found here: https://github.com/h2oai/h2oai-monitor/blob/master/Jenkinsfile#L129-L137

mwysokin (approved):
LGTM 🖖 

Sent you a link to one of our jenkinsfiles if you wanted to use only flags instead of env variables.

Jenkinsfile
achraf-mer: does it show in jenkins console logs? I thought it did so I used env vars

mmalohlava:
@achraf-mer do you target 1.1.8.3 branch? Isn't better to go directly into master?

achraf-mer:
> @achraf-mer do you target 1.1.8.3 branch? Isn't better to go directly into master?

@mmalohlava I can merge `master` after this gets merged into 1.1.8.3 just out of consistency since this is the release used by DAI 1.10.1.2, and we'll have to respect the order so `master` image can use the `latest` tag in the docker repositories.

Jenkinsfile
achraf-mer: I did as you suggested but had to revert because of errors, in order to debug error will have to turn on console logs, so I think best to have it with env vars.

mwysokin (approved):
LGTM 🖖

PR #281, opened 2022-01-05
Dynamically set capabilities endpoint based on configuration

nkpng2k:
Small improvement to `/model/capabilities` endpoint. No longer hardcoded. Pulls configuration from the scorer and uses that to set the response.


nkpng2k:
resolves https://github.com/h2oai/dai-deployment-templates

local-rest-scorer/src/main/java/ai/h2o/mojos/deploy/local/rest/controller/ModelsApiController.java
```diff
@@ -116,4 +118,26 @@ public ModelsApiController(MojoScorer scorer, SampleRequestBuilder sampleRequest
...
+        capabilityTypes.add(CapabilityType.CONTRIBUTION_ORIGINAL);
+        break;
+      case TRANSFORMED:
+        capabilityTypes.add(CapabilityType.CONTRIBUTION_TRANSFORMED);
+        break;
+      case NONE:
+      default:
+        break;
+    }
+    return capabilityTypes;
```
orendain, line 141: 100% fine and no issues, just providing the below to share style knowledge and some common tips:

- (naming convention) Since method returns and doesn't set member field, consider renaming method from `setSup...` to something like `assembleSupp..` or some other synonym.
- (reducing reliance/risk on specific instance initialization-order) Can reduce reliance on initialized instance variable by passing in `value of scorer.getEnabl..()` instead of looking it up from inside the method.  Would also free method up to become static.
- (readability) Returning immediately, instead of breaking, often allows reader to not finish reading method or keep state in their head as they scroll.  Similarly, can have each `case` return the full list outright without building (e.g., `List.of(..., ..., ....)`.  Means a few extra words of code, but is explicit, immutable, and no building logic involved.

orendain (approved):
LGTM - awesome!

local-rest-scorer/src/test/java/ai/h2o/mojos/deploy/local/rest/controller/ModelsApiControllerTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+  @BeforeAll
+  static void setup() {
+    System.setProperty("mojo.path", "src/test/resources/multinomial-pipeline.mojo");
+    mockDummyPipeline();
+  }
+
+  private static void mockDummyPipeline() {
+    MojoPipeline dummyPipeline =
+        new DummyPipeline(TEST_UUID, MojoFrameMeta.getEmpty(), MojoFrameMeta.getEmpty());
```
orendain, line 48: I wonder if our static mock troubles would be fixed if we moved the `loadMojoPipelineFromFile()` call in `MojoScorer.java` from static to the constructor.

local-rest-scorer/src/test/java/ai/h2o/mojos/deploy/local/rest/controller/ModelsApiControllerTest.java
```diff
@@ -0,0 +1,174 @@
...
+class ModelsApiControllerTest {
+  private static final String MOJO_PIPELINE_PATH = "src/test/resources/multinomial-pipeline.mojo";
+  private static final String TEST_UUID = "TEST_UUID";
+
+  @Mock private SampleRequestBuilder sampleRequestBuilder;
+
+  @BeforeAll
+  static void setup() {
+    System.setProperty("mojo.path", "src/test/resources/multinomial-pipeline.mojo");
+    mockDummyPipeline();
```
orendain, line 43: Doesn't look like the static mock session was closed.  Not a big deal since the class is small and each test relies on this.

But for thoroughness, I'll share a couple of ways through this:

1. Perform staticmock as part of a `try()` at the place it's needed, which auto cleans up the mock after use.  [Example.](https://github.com/h2oai/h2oai-serving/blob/2f9c0b8590ff5800cc10718769d4f023bdcf958a/common/src/test/java/ai/h2o/deploy/common/environment/EnvironmentsTest.java#L55)
2. Save the mock to a static variable and close the session as part of a `@AfterAll`.

local-rest-scorer/src/test/java/ai/h2o/mojos/deploy/local/rest/controller/ModelsApiControllerTest.java
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  /** Dummy pipeline {@link MojoPipeline} just to mock the static methods used inside scoring. */
+  private static class DummyPipeline extends MojoPipeline {
+    private final MojoFrameMeta inputMeta;
+    private final MojoFrameMeta outputMeta;
+
+    private DummyPipeline(String uuid, MojoFrameMeta inputMeta, MojoFrameMeta outputMeta) {
+      super(uuid, null, null);
+      this.inputMeta = inputMeta;
```
orendain, line 133: Take the following with a grain of salt, since it's usually not a good idea to mock objects you don't own:

Since you're using `DummyPipeline` in this test really just to get around one particular implementation issue (that we auto-load a pipeline as part of a `MojoScorer`), and aren't using this mock for anything really "mocky" ... you might be able to get away with defining a mock `MojoPipeline` (`@Mock private MojoPipeline dummyPipeline`)  rather than define a `DummyPipeline`.

Not sure what other issues might have come up, just sharing general tip.

local-rest-scorer/src/test/java/ai/h2o/mojos/deploy/local/rest/controller/ModelsApiControllerTest.java
```diff
@@ -0,0 +1,174 @@
...
+    System.setProperty("mojo.path", "src/test/resources/multinomial-pipeline.mojo");
+    mockDummyPipeline();
+  }
+
+  private static void mockDummyPipeline() {
+    MojoPipeline dummyPipeline =
+        new DummyPipeline(TEST_UUID, MojoFrameMeta.getEmpty(), MojoFrameMeta.getEmpty());
+    MockedStatic<MojoPipelineService> theMock = Mockito.mockStatic(MojoPipelineService.class);
+    theMock.when(() -> MojoPipelineService
+        .loadPipeline(new File(MOJO_PIPELINE_PATH))).thenReturn(dummyPipeline);
```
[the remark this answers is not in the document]
orendain: Circling back:  Ah, I see - the file isn't an actual MOJO.  Empty just to satisfy a `MojoScorer` assertion.  The comment about moving `loadMojoPIp...` from static to constructor seems even more relevant or potentially helpful.

Alternatively, if you don't want to commit an empty file for the purposes of satisfying that requirement, you can create a [random temporary file](https://devdocs.io/openjdk~11/java.base/java/nio/file/files#createTempFile(java.lang.String,java.lang.String,java.nio.file.attribute.FileAttribute...)) at runtime.

orendain (approved):
LGTM - tests seem valid 👍🏾.  Added some notes to help curb the workarounds or improve maintainability.

nkpng2k:
> LGTM - tests seem valid 👍🏾. Added some notes to help curb the workarounds or improve maintainability.

Thanks will fix and merge I think.

local-rest-scorer/src/main/java/ai/h2o/mojos/deploy/local/rest/controller/ModelsApiController.java
```diff
@@ -116,4 +120,25 @@ public ModelsApiController(MojoScorer scorer, SampleRequestBuilder sampleRequest
...
+    List<CapabilityType> capabilityTypes = new ArrayList<>();
+    capabilityTypes.add(CapabilityType.SCORE);
+    switch (enabledShapleyTypes) {
+      case ALL:
+        capabilityTypes.add(CapabilityType.CONTRIBUTION_ORIGINAL);
+        capabilityTypes.add(CapabilityType.CONTRIBUTION_TRANSFORMED);
+        return capabilityTypes;
+      case ORIGINAL:
+        capabilityTypes.add(CapabilityType.CONTRIBUTION_ORIGINAL);
+        return capabilityTypes;
```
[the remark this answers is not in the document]
nkpng2k: I can probably do `Arrays.asList` ... but `List.of` is Java 9+.

local-rest-scorer/src/test/java/ai/h2o/mojos/deploy/local/rest/controller/ModelsApiControllerTest.java
```diff
@@ -0,0 +1,174 @@
...
+    assertEquals(expectedCapabilities, response.getBody());
+  }
+
+  @Test
+  void verifyCapabilities_AllShapleyEnabled_ReturnsExpected() {
+    // Given
+    List<CapabilityType> expectedCapabilities = Arrays.asList(
+        CapabilityType.SCORE,
+        CapabilityType.CONTRIBUTION_ORIGINAL,
+        CapabilityType.CONTRIBUTION_TRANSFORMED);
```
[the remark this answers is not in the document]
nkpng2k: cant ... java8

local-rest-scorer/src/test/java/ai/h2o/mojos/deploy/local/rest/controller/ModelsApiControllerTest.java
nkpng2k: maybe... Seems like a bigger rework than I'm inclined to tackle at the moment.

local-rest-scorer/src/test/java/ai/h2o/mojos/deploy/local/rest/controller/ModelsApiControllerTest.java
nkpng2k: LOVE :heart_eyes: this definitely worked.

Issue #266, opened 2021-11-10
[Feature Request] Propagate mojo scorer images to public model registry for AWS sagemaker cpp scoring


achraf-mer:
### Request:
This request is a follow up to the issue in https://github.com/h2oai/h2oai/issues/26184
Ideally the image can expose the same API as the local runtime CLI.

arunaryasomayajula:
@Susankha any updates on this feature request? I recently had the customer ask for an update.

Susankha:
> @Susankha any updates on this feature request? I recently had the customer ask for an update.

@arunaryasomayajula I have sent PR - #284

PR #284, opened 2022-01-10
[Feature] AWS sagemaker hosted scorer for mojo - C++ runtime

Susankha:
This PR implements Sagemaker hosted mojo c++ scorer to fulfil feature request #266 with,

1. Docker file which using to build  docker container with all required libraries to run mojo c++ runtime.
2. Python REST API to serve scoring request coming through Sagemaker endpoint.

### DIFF
--- a/aws-sagemaker-hosted-scorer-cpp/Dockerfile
+++ b/aws-sagemaker-hosted-scorer-cpp/Dockerfile
@@ -1,16 +1,11 @@
 FROM python:3.8
 
-RUN apt-get update
+RUN apt-get -y update && pip3 install datatable \
+    flask-restful \
+    https://s3.amazonaws.com/artifacts.h2o.ai/releases/ai/h2o/daimojo/2.7.5/x86_64-centos7/daimojo-2.7.5-cp38-cp38-linux_x86_64.whl
 
-RUN pip3 install https://s3.amazonaws.com/artifacts.h2o.ai/releases/ai/h2o/daimojo/2.7.5/x86_64-centos7/daimojo-2.7.5-cp38-cp38-linux_x86_64.whl
-
-RUN pip3 install datatable
-
-RUN pip3 install flask-restful
-
-ENV DRIVERLESS_AI_LICENSE_FILE='/opt/ml/model/license.sig'
-
-ENV MOJO_FILE_PATH='/opt/ml/model/pipeline.mojo'
+ENV DRIVERLESS_AI_LICENSE_FILE='/opt/ml/model/license.sig' \
+    MOJO_FILE_PATH='/opt/ml/model/pipeline.mojo'
 
 RUN mkdir -p /opt/ml/code
 
--- a/aws-sagemaker-hosted-scorer-cpp/Dockerfile
+++ b/aws-sagemaker-hosted-scorer-cpp/Dockerfile
@@ -1,6 +1,8 @@
 FROM python:3.8
 
-RUN apt-get update && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && \
+    apt-get install -y libopenblas-dev && \
+    rm -rf /var/lib/apt/lists/*
 
 COPY requirements.txt /tmp/requirements.txt
 
--- a/aws-sagemaker-hosted-scorer-cpp/Dockerfile
+++ b/aws-sagemaker-hosted-scorer-cpp/Dockerfile
@@ -1,0 +1,21 @@
+FROM python:3.8
+
+RUN apt-get update
+
+RUN pip3 install https://s3.amazonaws.com/artifacts.h2o.ai/releases/ai/h2o/daimojo/2.7.5/x86_64-centos7/daimojo-2.7.5-cp38-cp38-linux_x86_64.whl
+
+RUN pip3 install datatable
+
+RUN pip3 install flask-restful
+
+ENV DRIVERLESS_AI_LICENSE_FILE='/opt/ml/model/license.sig'
+
+ENV MOJO_FILE_PATH='/opt/ml/model/pipeline.mojo'
+
+RUN mkdir -p /opt/ml/code
+
+COPY py/scorer/mojo-cpp-scorer.py /opt/ml/code
+
+WORKDIR /opt/ml/code
+
+ENTRYPOINT ["python", "mojo-cpp-scorer.py"]
--- a/aws-sagemaker-hosted-scorer-cpp/Dockerfile
+++ b/aws-sagemaker-hosted-scorer-cpp/Dockerfile
@@ -1,8 +1,10 @@
 FROM python:3.8
 
-RUN apt-get -y update && pip3 install datatable \
-    flask-restful \
-    https://s3.amazonaws.com/artifacts.h2o.ai/releases/ai/h2o/daimojo/2.7.5/x86_64-centos7/daimojo-2.7.5-cp38-cp38-linux_x86_64.whl
+RUN apt-get update && rm -rf /var/lib/apt/lists/*
+
+COPY requirements.txt /tmp/requirements.txt
+
+RUN pip install -r /tmp/requirements.txt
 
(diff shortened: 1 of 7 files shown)


aws-sagemaker-hosted-scorer-cpp/py/scorer/mojo_cpp_scorer.py
```diff
@@ -0,0 +1,124 @@
...
+logger = logging.getLogger(__name__)
+
+app = Flask(__name__)
+api = Api(app)
+
+
+class MojoPipeline(object):
+    _model = None
+    _instance = None
+    _lock = threading.Lock()
```
sh1ng, line 20: Why do you need this if you don't reuse `MojoPipline`?

aws-sagemaker-hosted-scorer-cpp/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM python:3.8
+
+RUN apt-get update
+
+RUN pip3 install https://s3.amazonaws.com/artifacts.h2o.ai/releases/ai/h2o/daimojo/2.7.5/x86_64-centos7/daimojo-2.7.5-cp38-cp38-linux_x86_64.whl
+
+RUN pip3 install datatable
+
+RUN pip3 install flask-restful
```
sh1ng, line 9: You can install multiple pkgs right away.

aws-sagemaker-hosted-scorer-cpp/py/scorer/mojo_cpp_scorer.py
Susankha: As this is the initial version, implemented like this because, modifications might be add depends on feedbacks. Hence I thought it could be reuse.

aws-sagemaker-hosted-scorer-cpp/Dockerfile
Susankha: sure, I will modify docker config accordingly.

nkpng2k:
@Susankha again, I would strongly suggest this go into the other repo https://github.com/h2oai/dai-deployment-examples here. It is much better suited for this pr.

Susankha:
> @Susankha I would probably suggest that this go here: https://github.com/h2oai/dai-deployment-examples , unless it is expected to respect the api defined in our `common/swagger.yaml` definitions.

@nkpng2k Thank you for pointing out this. I will discuss this with @mmalohlava and @achraf-mer

Susankha:
> @Susankha again, I would strongly suggest this go into the other repo https://github.com/h2oai/dai-deployment-examples here. It is much better suited for this pr.
> 
> cc @mmalohlava

I think this is not an example and this will build a docker image which include mojo c++ scorer libs and inference code to run on AWS sagemaker.

nkpng2k:
> 

I think there are 2 issues here:
1. this pr seems to be fully in python. Given this repo is nearly 100% java, I don't like the idea of polluting the project with different languages and build processes. 
2. Do we expect to maintain this for long periods of time or is this something that a user will take and branch/fork? If the latter then I'd still argue its a example (although the naming convention for the other repo is a little off anyways)

aws-sagemaker-hosted-scorer-cpp/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:3.8
+
+RUN apt-get -y update && pip3 install datatable \
```
mmalohlava, line 3: @Susankha better to have 2layers:

- first one is about system update (apt-get update, ...) - see Docker best practices - how to also cleanup `apt-get` cache to avoid spurious apt-get cache : https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run
- 2nd layer contains all pip install - i would even recommend to keep requirements file and do `pip install -r requirements.txt`

aws-sagemaker-hosted-scorer-cpp/README.md
```diff
@@ -0,0 +1,93 @@
+# DAI Deployment Template for Sagemaker Hosted C++ Scorer
+
+## Overview
+
+### Build Image
+
+Run the following command to build the docker image.
+
+```bash
+docker build -t <aws_account_id>.dkr.ecr.<region>.amazonaws.com/h2oai/sagemaker-hosted-scorer:<tag> .
```
mmalohlava, line 10: Can we get this as part of official gradle build. That image get pushed into our Vorvan registry?

aws-sagemaker-hosted-scorer-cpp/README.md
```diff
@@ -0,0 +1,93 @@
...
+Step 3:  Use the following curl command to test the container locally:
+
+```
+curl \
+    -X POST \
+    -H "Content-Type: application/json" \
+    -d @payload.json http://localhost:8080/invocations
+```
+
+payload.json:
```
mmalohlava, line 46: Is this payload compatible with swagger payload? if yes, we should mention it in documentation.

aws-sagemaker-hosted-scorer-cpp/README.md
```diff
@@ -0,0 +1,93 @@
+# DAI Deployment Template for Sagemaker Hosted C++ Scorer
```
mmalohlava, line 1: @Susankha do you want to write Teraform template for stack deployment? it could simplify the workflow.

aws-sagemaker-hosted-scorer-cpp/Dockerfile
Susankha: Done

aws-sagemaker-hosted-scorer-cpp/py/scorer/mojo_cpp_scorer.py
```diff
@@ -0,0 +1,124 @@
...
+        stypes=list(mojo.get_types().values())
+    )
+
+    d_frame = dt.fread(
+        text=tmp_frame.to_csv(),
+        columns=list(mojo.get_types().keys()),
+        na_strings=mojo.get_missing_values()
+    )
+
+    result = mojo.get_prediction(d_frame)
```
mmalohlava, line 115: you will need to do cbind with selected columns from `d_frame`.

mmalohlava:
What is missing is some test to fix the API expectations on request and response.

aws-sagemaker-hosted-scorer-cpp/py/scorer/mojo_cpp_scorer.py
Susankha: Done

aws-sagemaker-hosted-scorer-cpp/py/scorer/mojo_cpp_scorer.py
sh1ng: It's better to add it when you use it. For not it's just an extra code and complexity that adds nothing. Plus we need to test mojo more carefully for reuse cases.

sh1ng (changes requested):

arunaryasomayajula:
any updates @Susankha

aws-sagemaker-hosted-scorer-cpp/py/scorer/mojo_cpp_scorer.py
achraf-mer: This is fine, thread-safe singleton in case Flask is running with multi-threading.

sh1ng (commented):
Let's set `OMP_NUM_THREADS` based on cores count(`nproc/2`)

achraf-mer:
> Let's set `OMP_NUM_THREADS` based on cores count(`nproc/2`)

done here: https://github.com/h2oai/dai-deployment-templates/blob/mojo_cpp_scorer/aws-sagemaker-hosted-scorer-cpp/py/scorer/mojo_cpp_scorer.py#L61 I created a separate branch from where I can create a new build. @Susankha please feel free to adapt. Thanks

PR #300, opened 2022-04-27
[BUILD] Release 1.1.15

Susankha:


achraf-mer (commented):
Just a small note with regards to branching new releases, we usually keep the main branch up to date with latest code and then branch off, not the other way around, this is to avoid having a release branch with more commits than main branch, it just makes things easy to maintain. See https://github.com/h2oai/dai-deployment-templates#step-by-step
Thanks.

achraf-mer (approved):

PR #303, opened 2022-06-01
Address CVE-2019-10086 by adding dependency constraint to the beanutils versions that mojo2-runtime-impl expects

achraf-mer:
Mitigate CVE-2019-10086

(Mojo already uses the expected version at: https://github.com/h2oai/mojo2/pull/1209/files)

### DIFF
--- a/aws-lambda-scorer/lambda-template/build.gradle
+++ b/aws-lambda-scorer/lambda-template/build.gradle
@@ -5,6 +5,10 @@ dependencies {
     implementation project(':common:transform')
     implementation group: 'ai.h2o', name: 'mojo2-runtime-api'
     implementation group: 'ai.h2o', name: 'mojo2-runtime-impl'
+    constraints {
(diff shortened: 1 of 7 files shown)


aws-lambda-scorer/lambda-template/build.gradle
```diff
@@ -5,6 +5,10 @@ dependencies {
     implementation project(':common:transform')
     implementation group: 'ai.h2o', name: 'mojo2-runtime-api'
     implementation group: 'ai.h2o', name: 'mojo2-runtime-impl'
+    constraints {
+        // because https://app.snyk.io/vuln/SNYK-JAVA-COMMONSBEANUTILS-460111
+        compile group: 'commons-beanutils', name: 'commons-beanutils', version: '1.9.4'
```
this, line 10: @achraf-mer Did you check the final JARs? Because, sometimes `compil` time constraints doesn't affects the dependency versions in the final JAR

PR #314, opened 2022-07-06
[API] Introduces new "Score Media" API

nkpng2k:
Introduces api for image scoring in openapi 3.0 spec similar to https://github.com/h2oai/mlops-byom-images/pull/63

* moves original api spec to `v1` directory, but preserves build policies and paths (uses openapi v2.0 spec)
* introduces new `v2` api directory which implements openapi v3.0 spec
* introduces additional api endpoint to match above pr 
* bumps several dependencies including gradle (hence @achraf-mer why I'm tagging you)
* implementes empty `UNIMPLEMENTED` endpoint in `local-rest-scorer` 
* adds some tests.


achraf-mer (approved):
LGTM, once built we can run a scan to check.

common/swagger/v2/swagger.yaml
```diff
@@ -0,0 +1,71 @@
...
+                $ref: '../v1/swagger.yaml#/definitions/ScoreResponse'
+        '501':
+          description: Implementation not supported
+        '400':
+          description: Invalid payload
+components:
+  schemas:
+    ScoreMediaRequest:
+      allOf:
+        - $ref: '../v1/swagger.yaml#/definitions/ScoreRequest'
```
orendain, line 62: So as to not be tied to the evolution of the non-media v1 spec, could this and the response have its own explicit definition?

common/swagger/v2/swagger.yaml
```diff
@@ -0,0 +1,71 @@
...
+      operationId: getScoreMedia
+      requestBody:
+        content:
+          multipart/form-data:
+            schema:
+              type: object
+              properties:
+                payload:
+                  $ref: '#/components/schemas/ScoreMediaRequest'
+                files:
```
orendain, line 39: Does this mean that there must be a field labeled `payload` in the POST call?  What is the type of that field?  Is it an object itself, or a JSON string?

Would it make sense to flatten the structure so that those fields are on the same level as `files`?

orendain (changes requested):
Requested changes was for explicit definition of objects in the API (untied from v1 API), but had some other questions.

Thank you @nkpng2k !

local-rest-scorer/src/main/java/ai/h2o/mojos/deploy/local/rest/controller/ModelsApiControllerV2.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+@Controller
+public class ModelsApiControllerV2 implements ModelApi {
+
+  private static final Logger log = LoggerFactory.getLogger(ModelsApiControllerV2.class);
+
+  // TODO (@nkpng2k): https://github.com/OpenAPITools/openapi-generator/issues/1646
+  // Generated code does not work properly despite proper spec.
+  @Override
+  public ResponseEntity<ScoreResponse> getScoreMedia(
```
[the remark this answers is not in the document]
nkpng2k: oh oops... forgot to delete that comment :sweat_smile:

local-rest-scorer/src/main/java/ai/h2o/mojos/deploy/local/rest/converters/ScoreMediaRequestConverter.java
```diff
@@ -0,0 +1,13 @@
...
+import ai.h2o.mojos.deploy.common.rest.v2.model.ScoreMediaRequest;
+import com.google.gson.Gson;
+import org.springframework.core.convert.converter.Converter;
+
+public class ScoreMediaRequestConverter implements Converter<String, ScoreMediaRequest> {
+
+  @Override
+  public ScoreMediaRequest convert(String input) {
+    return new Gson().fromJson(input, ScoreMediaRequest.class);
+  }
```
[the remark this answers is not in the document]
nkpng2k: slightly different. The `transformers` in the above are related to data transformations as opposed to http request transformations.... I suppose we could have some common http converters, but this is a springboot specific thing so unrelated to the code in `transformers`

common/swagger/v2/swagger.yaml
nkpng2k: tbh I debated this as well... in fact, I even considered having a single request spec that both apis inherit from, because in reality it makes much more sense to have the requests and responses be unified and evolve independently of either api spec. Like each api should be dependent on the requests and responses themselves and while the apis can evolve they would be constrained by the requests/responses such that there is some level of consistency.

common/swagger/v2/swagger.yaml
```diff
@@ -0,0 +1,71 @@
...
+servers:
+  - url: '{protocol}://127.0.0.1:{port}/'
+    variables:
+      protocol:
+        enum:
+          - http
+          - https
+        default: http
+      port:
+        default: '8080'
```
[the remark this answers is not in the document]
nkpng2k: I think `url` can just be `/` I'll try it out.

common/swagger/v2/swagger.yaml
```diff
@@ -0,0 +1,71 @@
+openapi: '3.0.0'
+info:
+  description: >-
+    This is an extension of the REST API for scoring from H2O. This API is
+    intended to be used for scoring within H2O.ai MLOps.
```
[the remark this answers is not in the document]
nkpng2k: :+1:

common/swagger/v2/swagger.yaml
nkpng2k: yes the request would take the format:
```python
res = requests.post("http://localhost:8080/model/score/media", data={"payload": json.dumps(req)}, files=files)
```
using the requests package in python.

Sadly, no this is not possible. Form data does not support complex objects like `list of list` therefore it is not possible to define the payload as independent form fields.

nkpng2k: @orendain

nkpng2k:
@orendain I added new model specifications for v2 api, so now each of the apis are now independent of each other... although at the moment that means a ton of duplicated "code"

gradle/mixins/dependencies.gradle
```diff
@@ -50,6 +50,7 @@ dependencyManagement {
         dependency group: 'ai.h2o', name: 'sparkling-water-scoring_2.12', version: sparklingWaterVersion
         dependency group: 'com.typesafe', name:'config', version: configVersion
         dependency group: 'uk.org.webcompere', name: 'system-stubs-jupiter', version: jupiterSystemStubsVersion
-
+        dependency group: 'com.squareup.okhttp', name: 'okhttp', version: okhttpVersion
+        dependency group: 'com.squareup.okhttp', name: 'logging-interceptor', version: okhttpVersion
```
orendain, line 54: Didn't notice this the first time.  QQ: why are these needed?

orendain (commented):
Performance considerations aside (e.g., whether an underlying scorer can calculate contribution with less compute during scoring than if done separately), WDYT about removing the contributions-related bits from this endpoint?

- There had been some multiple discussions about whether it makes sense to keep scoring and shap endpoints separated.  I can't recall/find arguments for keeping them combined at an API level.
- This v2 API is not necessarily meant to be backward compatible with the v1 Scoring API. Adopting the API from v1 Scoring API makes this API an extension and not be able to serve as the next evolution.  For example, if when deprecating v1, the decision is to separate the concerns above, the endpoint here becomes a conflict.
- Granted, this may raise a feature request to support shapley values for image scoring, which would also come baked if the API looked identical to v1.
- If not, then perhaps we should mark this API not as v2 but as a v1 extension?  And reserve v2 for when APIs are evolved.

WDYT?

---

Is it possible to separate the implementation from the spec?  ~20-25 files are related to the new spec (and moving things around), and ~5 to the implementation in the local rest scorer.  That way, we also hold off from tying the scorer to changes in v2 API, especially if there aren't any plans to do anything other than throw unimplemented.

common/swagger/v2/swagger.yaml
nkpng2k: ummmm... @orendain just to be clear `shapley contributions` != `image/media scoring`

sure we could flatten the payload, I guess the question becomes, do we want to validate a single string, or multiple different strings, in the backend, you'd end up having to parse 2 strings using some version of `json.loads()` to get fields which is an array of strings, and rows which is an array of arrays of strings. Personally, I don't like this... like from client call perspective, it increases the amount of work required to construct the proper payload request, and does not (imho) significantly improve the ability to understand or utilize the api.

nkpng2k:
offline discussion with @orendain:
1. extract implementation from api spec pr
2. use more rest compliant endpoint. ex `/model/mediaScore` (? maybe not camel case... check restful spec)
3. not v2, v1exp/v1ext

nkpng2k:
@orendain completed changes as per offline discussion, also reverted all changes associated with implementation in `local-rest-scorer`

orendain (approved):
Clean new experimental API without affecting any existing binaries!

LGTM. Thank you @nkpng2k!

PR #317, opened 2022-08-11
switching to mojo2-runtime version 2.7.10

pkozelka:
https://github.com/h2oai/mojo2/pull/1534


orendain:
Heads up that v`2.7.11` is now out.

PR #318, opened 2022-08-24
Invert remaining v1exp API terms `ScoreMedia` to `MediaScore`.

orendain:
The endpoint, `media-score` was modified to properly represent a noun, though a few terms in the API definition were missing.  This PR makes those remaining changes.


nkpng2k (approved):
Lgtm

PR #323, opened 2022-12-12
Don't log failed score request content by default

g-eoj:
Moves local rest server score request content logging to DEBUG. Score request content is only logged when scoring fails. The content of the request may be considered sensitive and thus we shouldn't log it by default.

Partially addresses https://github.com/h2oai/model-manager/issues/1570.


orendain (approved):
LGTM

PR #326, opened 2023-02-28
Add `runtimeOnly` dependency on MOJO2 libraries with additional transformer support.

orendain:
> **Note**
> This is not the ideal implementation.  Control of expected logger is lost and logs look different.  Leaving as a draft.  Will edit this description with more info later.


jakubhava:
Closing as stale

PR #327, opened 2023-03-29
Add scoring request transformer to sanitize bool literal string

jackjii79:
Part of https://github.com/h2oai/model-manager/issues/1816
https://github.com/h2oai/model-manager/issues/1816#issuecomment-1487716663

### DIFF
--- a/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
+++ b/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
@@ -57,7 +57,7 @@ public class MojoScorer {
   private final ContributionRequestToMojoFrameConverter contributionRequestConverter;
   private final MojoPipelineToModelInfoConverter modelInfoConverter;
   private final CsvToMojoFrameConverter csvConverter;
-  private final ScoreRequestTransformer scoreRequestTransformer = new ScoreRequestTransformer();
+  private final ScoreRequestTransformer scoreRequestTransformer;
 
   /**
    * MojoScorer class initializer, requires below parameters.
@@ -73,12 +73,14 @@ public class MojoScorer {
       ContributionRequestToMojoFrameConverter contributionRequestConverter,
       MojoFrameToContributionResponseConverter contributionResponseConverter,
       MojoPipelineToModelInfoConverter modelInfoConverter,
+      ScoreRequestTransformer scoreRequestTransformer,
       CsvToMojoFrameConverter csvConverter) {
     this.scoreRequestConverter = scoreRequestConverter;
     this.scoreResponseConverter = scoreResponseConverter;
     this.contributionRequestConverter = contributionRequestConverter;
     this.contributionResponseConverter = contributionResponseConverter;
     this.modelInfoConverter = modelInfoConverter;
+    this.scoreRequestTransformer = scoreRequestTransformer;
     this.csvConverter = csvConverter;
 
     this.enabledShapleyTypes = ShapleyLoadOption.fromEnvironment();
--- a/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
+++ b/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
@@ -57,6 +57,7 @@ public class MojoScorer {
   private final ContributionRequestToMojoFrameConverter contributionRequestConverter;
   private final MojoPipelineToModelInfoConverter modelInfoConverter;
   private final CsvToMojoFrameConverter csvConverter;
+  private final ScoreRequestTransformer scoreRequestTransformer = new ScoreRequestTransformer();
 
   /**
    * MojoScorer class initializer, requires below parameters.
@@ -93,6 +94,7 @@ public class MojoScorer {
    * @return response {@link ScoreResponse}
    */
   public ScoreResponse score(ScoreRequest request) {
+    scoreRequestTransformer.accept(request, getModelInfo().getSchema().getInputFields());
     MojoFrame requestFrame = scoreRequestConverter
             .apply(request, pipeline.getInputFrameBuilder());
     MojoFrame responseFrame = doScore(requestFrame);
--- a/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformer.java
+++ b/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformer.java
@@ -1,0 +1,49 @@
+package ai.h2o.mojos.deploy.common.transform;
+
+import ai.h2o.mojos.deploy.common.rest.model.DataField;
+import ai.h2o.mojos.deploy.common.rest.model.Row;
+import ai.h2o.mojos.deploy.common.rest.model.ScoreRequest;
+import com.google.common.collect.ImmutableMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * Transform scoring request rows,
+ * specifically, convert boolean literal string into 1, 0 respectively.
+ */
+public class ScoreRequestTransformer implements BiConsumer<ScoreRequest, List<DataField>> {
+
+  private static final Logger logger = LoggerFactory.getLogger(ScoreRequestTransformer.class);
+
+  @Override
+  public void accept(ScoreRequest scoreRequest, List<DataField> dataFields) {
+    Map<String, DataField> dataFieldMap = dataFields.stream().collect(ImmutableMap.toImmutableMap(DataField::getName, Function.identity()));
+    scoreRequest.setRows(transformRow(scoreRequest.getFields(), scoreRequest.getRows(), dataFieldMap));
+  }
+
+  List<Row> transformRow(List<String> fields, List<Row> rows, Map<String, DataField> dataFields) {
+    return rows.stream().map(row -> {
+      List<String> transformData = IntStream.range(0, row.size()).mapToObj(
+          fieldIdx -> {
+            String colName = fields.get(fieldIdx);
+            String origin = row.get(fieldIdx);
+            String sanitizeValue = Utils.sanitizeBoolean(origin, dataFields.get(colName).getDataType());
+            if (!sanitizeValue.equals(origin)) {
+              logger.info("Value '{}' parsed as '{}'", origin, sanitizeValue);
+            }
+            return sanitizeValue;
+          }
+      ).collect(Collectors.toList());
+      Row transformedRow = new Row();
+      transformedRow.addAll(transformData);
+      return transformedRow;
+    }).collect(Collectors.toList());
+  }
+}
--- a/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformer.java
+++ b/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformer.java
@@ -47,7 +47,7 @@ public class ScoreRequestTransformer implements BiConsumer<ScoreRequest, List<Da
                   origin, dataFields.get(colName).getDataType()
               );
               if (!sanitizeValue.equals(origin)) {
-                logger.info("Value '{}' parsed as '{}'", origin, sanitizeValue);
+                logger.debug("Value '{}' parsed as '{}'", origin, sanitizeValue);
               }
               return sanitizeValue;
             } else {
--- a/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformer.java
+++ b/common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformer.java
@@ -4,8 +4,6 @@ import ai.h2o.mojos.deploy.common.rest.model.DataField;
 import ai.h2o.mojos.deploy.common.rest.model.Row;
 import ai.h2o.mojos.deploy.common.rest.model.ScoreRequest;
 import com.google.common.collect.ImmutableMap;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.util.List;
 import java.util.Map;
@@ -13,6 +11,8 @@ import java.util.function.BiConsumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Transform scoring request rows,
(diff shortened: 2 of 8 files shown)


common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/Utils.java
```diff
@@ -23,4 +24,22 @@ public static void copyResultFields(MojoFrame mojoFrame, List<Row> outputRows) {
       }
     }
   }
+
+  /**
+   * Sanitize boolean string literal values true / false (case insensitive) into 1 / 0 respectively.
+   * @return sanitized string.
+   */
+  public static String sanitizeBoolean(String value, DataField.DataTypeEnum dataType) {
+    if (
```
achraf-mer, line 33: should it be false is empty?
also, how about `T`/`F`?

common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformerTest.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @Test
+  void transform_NonBooleanLiteral_Unchanged() {
+    // Given
+    ScoreRequest scoreRequest = new ScoreRequest();
+    scoreRequest.setFields(Collections.singletonList("test"));
+    List<Row> rows = new ArrayList<>(Arrays.asList(new Row(), new Row()));
+    rows.get(0).addAll(Collections.singletonList("dumb1"));
+    rows.get(1).addAll(Collections.singletonList("dumb2"));
```
achraf-mer, line 64: nit: can use some string other than 'dumb'

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/Utils.java
jackjii79: > should it be false is empty?

We try not to assume how to handle missing values, the responsibility moves down to the MOJO pipeline as the processing of missing values should be consistent with C++ MOJO runtime as well, where we just leave the input to be whatever they were in this case.

> also, how about T/F?

the assumption is when data type is `float` and input is `true / false` case insensitive then we casting into `1 / 0` respectively,  because scoring pipeline and mojo pipeline cast all boolean columns as float type. in my impression, `T / F` is too fishy to make any assumption about the original column, where it could also mean some categorical features.

common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformerTest.java
jackjii79: thx @achraf-mer ! good idea

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/Utils.java
g-eoj: @achraf-mer we're trying to accomdate DAIs behavior. How does DAI handle `T/F`? Are there any other boolean conversions made by DAI?

common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformerTest.java
```diff
@@ -0,0 +1,80 @@
...
+import ai.h2o.mojos.deploy.common.rest.model.Row;
+import ai.h2o.mojos.deploy.common.rest.model.ScoreRequest;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class ScoreRequestTransformerTest {
```
aravind-viswanathan, line 15: Considering that there is a case wherein the field is not present. Should we have a test case covering that scenario as well.

aravind-viswanathan (commented):
Need a clarification on the possibility of us skipping boolean fields if it is not part of our known schema.

common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformerTest.java
jackjii79: good idea ! updated https://github.com/h2oai/dai-deployment-templates/pull/327/commits/844043a958779dce72c770bd3d8e72d72ea41959

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformer.java
```diff
@@ -0,0 +1,62 @@
...
+    );
+  }
+
+  List<Row> transformRow(List<String> fields, List<Row> rows, Map<String, DataField> dataFields) {
+    return rows.stream().map(row -> {
+      List<String> transformData = IntStream.range(0, row.size()).mapToObj(
+          fieldIdx -> {
+            String colName = fields.get(fieldIdx);
+            String origin = row.get(fieldIdx);
+            if (dataFields.containsKey(colName)) {
```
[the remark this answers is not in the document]
jackjii79: We have to stick to the schema, it is the schema that tells us whether or not we should sanitize or not, the assumption is any field with `float` type with literal string like `true` or `false` case insensitive then we cast into binary so to be consistent with the behavior of DAI, if the received field name is not part of the input schema then we simply ignore it.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformer.java
```diff
@@ -0,0 +1,62 @@
...
+        dataFields.stream().collect(
+            ImmutableMap.toImmutableMap(DataField::getName, Function.identity()
+            )
+        );
+    scoreRequest.setRows(
+        transformRow(scoreRequest.getFields(), scoreRequest.getRows(), dataFieldMap)
+    );
+  }
+
+  List<Row> transformRow(List<String> fields, List<Row> rows, Map<String, DataField> dataFields) {
```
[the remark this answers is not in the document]
jackjii79: no reason not to do so, i probably overlooked, good catch ! https://github.com/h2oai/dai-deployment-templates/pull/327/commits/844043a958779dce72c770bd3d8e72d72ea41959

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformer.java
```diff
@@ -0,0 +1,62 @@
...
+          fieldIdx -> {
+            String colName = fields.get(fieldIdx);
+            String origin = row.get(fieldIdx);
+            if (dataFields.containsKey(colName)) {
+              String sanitizeValue = Utils.sanitizeBoolean(
+                  origin, dataFields.get(colName).getDataType()
+              );
+              if (!sanitizeValue.equals(origin)) {
+                logger.info("Value '{}' parsed as '{}'", origin, sanitizeValue);
+              }
```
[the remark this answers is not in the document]
jackjii79: `mojo2` runtime expose the same message on `INFO` level when transform boolean string literal into numeric value,  the only difference is `mojo2` can only handle lowercase or uppercase, because we doing something similar to `mojo2` which would has the equivalent of impact to the JAVA runtime, so i just keep the logging behavior consistent with existing convention.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -57,6 +57,7 @@ public class MojoScorer {
   private final ContributionRequestToMojoFrameConverter contributionRequestConverter;
   private final MojoPipelineToModelInfoConverter modelInfoConverter;
   private final CsvToMojoFrameConverter csvConverter;
+  private final ScoreRequestTransformer scoreRequestTransformer = new ScoreRequestTransformer();
```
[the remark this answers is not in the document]
jackjii79: you are absolutely right, initial i trying to minimize the scope of changes so to leave the constructor unchanged since it is also used in different libraries but it is not a good practice, updated https://github.com/h2oai/dai-deployment-templates/pull/327/commits/844043a958779dce72c770bd3d8e72d72ea41959

jackjii79:
Thanks to the review @aravind-viswanathan ! Made some updates, PTAL when you have a chance 🖕

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformer.java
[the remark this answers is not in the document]
g-eoj: @jackjii79 if this is logging for every row that is not a good idea. The logs will get spammed.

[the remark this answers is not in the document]
jackjii79: ok, it is indeed per row, changed to debug level https://github.com/h2oai/dai-deployment-templates/pull/327/commits/ba9ccdd75cd060f683f38a6a6642006be8459994

aravind-viswanathan (approved):
LGTM

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/Utils.java
```diff
@@ -23,4 +24,22 @@ public static void copyResultFields(MojoFrame mojoFrame, List<Row> outputRows) {
...
+        dataType.equals(DataField.DataTypeEnum.FLOAT32)
+            || dataType.equals(DataField.DataTypeEnum.FLOAT64)
+    ) {
+      if (value.matches("(?i)true")) {
```
rasika, line 37: nit: I would go for `if ("true".equalsIgnoreCase(value)) {` for safe handling of null values and avoid using inline regexes.

jackjii79: good catch ! thx @rasika , wonder why i did not come up with this in the first place.. updated https://github.com/h2oai/dai-deployment-templates/pull/327/commits/ece8b620f6342a4f70e1c20a32c368ee06738b57

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/Utils.java
```diff
@@ -23,4 +24,22 @@ public static void copyResultFields(MojoFrame mojoFrame, List<Row> outputRows) {
...
+  /**
+   * Sanitize boolean string literal values true / false (case insensitive) into 1 / 0 respectively.
+   * @return sanitized string.
+   */
+  public static String sanitizeBoolean(String value, DataField.DataTypeEnum dataType) {
+    if (
+        dataType.equals(DataField.DataTypeEnum.FLOAT32)
+            || dataType.equals(DataField.DataTypeEnum.FLOAT64)
+    ) {
+      if (value.equalsIgnoreCase("true")) {
```
achraf-mer, line 37: best to compare against "true" and not `value`, to avoid any potential npe.

jackjii79: thx @achraf-mer , see my response:
- our API is built on the swagger schema to parse the request payload. And according to the schema, `Row` is an array of String https://github.com/h2oai/dai-deployment-templates/blob/692de21b35d516e9da8b53d36dc546fcaf16264c/common/swagger/v1/swagger.yaml#L180-L183
Even if we get around with the issue by reversing the order, that is invalid request, even worse that could hide the real problem, if it is going to be a unchecked exception, the best thing to handle is just to fail fast

 - i doubt it would be possible to for our API handler to receive a null value, as all requests coming through monitor proxy which will do the validation work to see the requests are subject to the schema or not, it will exclude all invalid requests, just show you an example in below:

send a `null` value in payload
```
jinghan@JHan-MBP16 h2o % curl -X POST -H "Content-Type: application/json" -d @- https://model.jh-test-1.mlops-dev.h2o.ai/f48f64b1-c3ef-410c-9df7-40098b679563/model/score << EOF
{
  "fields": [
    "timestamp",
    "count",
    "visibility",
    "temperature",
    "windSpeed",
    "humidity",
    "pressure",
    "substation"
  ],
  "rows": [
    [
      null,
      "0",
      "0",
      "0",
      "0",
      "0",
      "0",
      "0"
    ]
  ]
}
EOF
```

exception from monitor proxy
```
2023-04-06 17:32:49,162 [ERROR] akka.actor.ActorSystemImpl - Error during processing of request: 'Expected String as JsString, but got null'. Completing with 500 Internal Server Error response. To change default exception handling behavior, provide a custom ExceptionHandler.                                                                                                        │
│ spray.json.DeserializationException: Expected String as JsString, but got null                                                                                                                                                                                                                                                                                                             │
│     at spray.json.package$.deserializationError(package.scala:23)                                                                                                                                                                                                                                                                                                                          │
│     at spray.json.ProductFormats.fromField(ProductFormats.scala:63)                                                                                                                                                                                                                                                                                                                        │
│     at spray.json.ProductFormats.fromField$(ProductFormats.scala:51)                                                                                                                                                                                                                                                                                                                       │
│     at ai.h2o.monitor.proxy.localrestscorer.LocalRestScorerFormat$.fromField(LocalRestScorerFormat.scala:7)                                                                                                                                                                                                                                                                                │
│     at spray.json.ProductFormatsInstances$$anon$4.read(ProductFormatsInstances.scala:109)                                                                                                                                                                                                                                                                                                  │
│     at spray.json.ProductFormatsInstances$$anon$4.read(ProductFormatsInstances.scala:95)                                                                                                                                                                                                                                                                                                   │
│     at ai.h2o.monitor.proxy.localrestscorer.LocalRestScorerRequest.<init>(LocalRestScorerRequest.scala:11)                                                                                                                                                                                                                                                                                 │
│     at ai.h2o.monitor.proxy.localrestscorer.LocalRestScorer.newScoringRequest(LocalRestScorer.scala:63)                                                                                                                                                                                                                                                                                    │
│     at ai.h2o.monitor.proxy.localrestscorer.LocalRestScorer.newScoringRequest(LocalRestScorer.scala:36)                                                                                                                                                                                                                                                                                    │
│     at ai.h2o.monitor.proxy.ProxyRoute.$anonfun$monitorEndpoint$1(ProxyRoute.scala:34) 
```

achraf-mer: it does not matter if it can get to to code or not, at least for now, it is generally a good practice to compare with reference to the known string than the variable.

jackjii79: @achraf-mer ok, i see your point and i m not insist on not changing that, just fyi it is absolutely violate the request schema if it is a `null` value and with or without monitor proxy, the unchecked exception would be a normal reaction at API layer.
updated 
https://github.com/h2oai/dai-deployment-templates/pull/327/commits/7c9b0ec62ffe42650fa4282eabaed1dfd40f27ce

common/transform/src/test/java/ai/h2o/mojos/deploy/common/transform/ScoreRequestTransformerTest.java
jackjii79: https://github.com/h2oai/dai-deployment-templates/pull/327/commits/0a60f7cf243fd393b8e542e707de53d68d235132

rasika (approved):

PR #328, opened 2023-03-30
Make LOGLEVEL configurable in rest scorer

jackjii79:
Address https://github.com/h2oai/model-manager/issues/1886

By toggle `LOGLEVEL`, logging is configurable
For instance, when `LOGLEVEL` is `debug`
```
│ 2023-03-30 17:58:05.326  INFO 1 --- [nio-8080-exec-9] a.h.m.d.l.r.c.ModelsApiController        : Got scoring request                                                                                                                                                                                                                                                                       │
│ 2023-03-30 17:58:05.331 DEBUG 1 --- [nio-8080-exec-9] a.h.m.d.common.transform.MojoScorer      : Input has 1 rows, 23 columns: [SEX, EDUCATION, MARRIAGE, AGE, PAY_1, PAY_2, PAY_3, PAY_4, PAY_5, PAY_6, BILL_AMT1, BILL_AMT2, BILL_AMT3, BILL_AMT4, BILL_AMT5, BILL_AMT6, PAY_AMT1, PAY_AMT2, PAY_AMT3, PAY_AMT4, PAY_AMT5, PAY_AMT6, DEFAULT_PAYMENT_NEXT_MONTH]                         │
│ 2023-03-30 17:58:05.335 DEBUG 1 --- [nio-8080-exec-9] a.h.m.d.common.transform.MojoScorer      : Response has 1 rows, 1 columns: [LIMIT_BAL]  
```

### DIFF
--- a/local-rest-scorer/src/main/resources/application.properties
+++ b/local-rest-scorer/src/main/resources/application.properties
@@ -1,2 +1,1 @@
-logging.level.ai.h2o.mojos.deploy.common.transform.MojoScorer=${LOGLEVEL}
-logging.level.ai.h2o.mojos.deploy.local.rest.controller.ModelsApiController=${LOGLEVEL}
+logging.level.org.springframework.boot=${LOGLEVEL}
--- a/local-rest-scorer/src/main/resources/application.properties
+++ b/local-rest-scorer/src/main/resources/application.properties
@@ -1,0 +1,2 @@
+logging.level.ai.h2o.mojos.deploy.common.transform.MojoScorer=${LOGLEVEL}
+logging.level.ai.h2o.mojos.deploy.local.rest.controller.ModelsApiController=${LOGLEVEL}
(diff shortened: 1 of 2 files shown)


g-eoj (commented):
Is this picking up an env var name `LOGLEVEL`? Is that a conventional name for the var or should it be a rest scorer specific name?

jackjii79:
> Is this picking up an env var name `LOGLEVEL`? Is that a conventional name for the var or should it be a rest scorer specific name?

Here is the ref in springBoot : https://docs.spring.io/spring-boot/docs/2.1.13.RELEASE/reference/html/boot-features-logging.html#boot-features-custom-log-levels

It is points to the logger name we defined, in this case, it is `MojoScorer` and api controller becasue the way we create logger is by its class name like 

```
LoggerFactory.getLogger(class.name)
```

g-eoj:
I'm asking why is the env var called `LOGLEVEL` instead of something specific to h2o like `LOGGING_LEVEL_H2O_REST_SERVER`.

jackjii79:
> I'm asking why is the env var called `LOGLEVEL` instead of something specific to h2o like `LOGGING_LEVEL_H2O_REST_SCORER`.

I think it is easier for user to overwrite the loglevel for whatever runtime they need with a uniform name to control the log level, `LOGLEVEL` is going to be used across all our runtime, i dont think there is a need to specialize the naming of the variable

g-eoj (approved):
Thanks!

jackjii79:
To give more context, this is the way that BE been doing for a while, but functionally it never seems working
https://github.com/h2oai/h2oai-serving/blob/245a8e0baa3a0aa8433a89b2fdd65e036eaf0b5b/deployment-library/src/main/java/ai/h2o/deploy/kubernetes/kit/RuntimeKit.java#L130

PR #329, opened 2023-03-31
allow MODEL_ID to override pipeline uuid

jackjii79:
Can not repro https://github.com/h2oai/model-manager/issues/1820 but this PR will make sure scoring API always pull from overwritten `MODEL_ID` passed through deployment


g-eoj (approved):
LGTM. Thanks!

PR #330, opened 2023-03-31
Handle scroing exception at API layer

jackjii79:
Part of https://github.com/h2oai/model-manager/issues/1891

## demo
```
jinghan@JHan-MBP16 h2o % curl -X POST -H "Content-Type: application/json" -d @- https://model.jh-test-1.mlops-dev.h2o.ai/3f49b90c-27a4-40bb-acd7-4b12d84f8367/model/score << EOF
{
  "fields": [
    "timestamp",
    "count",
    "visibility",
    "temperature",
    "windSpeed",
    "humidity",
    "pressure",
    "substation"
  ],
  "rows": [
    [
      "text",
      "this is a test",
      "0",
      "0",
      "0",
      "0",
      "0",
      "0"
    ]
  ]
}
EOF
Illegal Server response: {"detail":"Failed scoring request due to: java.lang.NumberFormatException: For input string: \"this is a test\""}
```

PR #331, opened 2023-04-07
Default exception handler should always capture error

jackjii79:


jackjii79:
in favor of https://github.com/h2oai/dai-deployment-templates/pull/330 and https://github.com/h2oai/dai-deployment-templates/pull/333

PR #335, opened 2023-04-18
build: Release 1.1.19

mmalohlava:
- build: Upgrade dependencies
- build: 🚨 TMP: added explicit dependency on opencsv 🚨
- build: Release 1.1.19
- build: Next devel cycle 1.1.20-SNAPSHOT


mmalohlava:
@orendain @jackjii79 @mwysokin i will go ahead with this PR even though it is not fully green (one red is old CodeQL, the second one is misconfigured snyk (needs to be configured for right Gradle configurations else reporting everything including test code))

orendain (commented):
👍🏾

PR #337, opened 2023-04-25
Support prediction interval

jackjii79:
https://github.com/h2oai/model-manager/issues/1664


common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/Utils.java
```diff
@@ -42,4 +47,27 @@ public static String sanitizeBoolean(String value, DataField.DataTypeEnum dataTy
...
+  public static boolean isPredictionIntervalEnabled() {
+    return isPredictionIntervalEnabledFromEnv() || isPredictionIntervalEnabledFromProperty();
+  }
+
+  private static boolean isPredictionIntervalEnabledFromEnv() {
+    try {
+      return Boolean.getBoolean(PREDICTION_INTERVAL_ENABLE_PROPERTY);
+    } catch (Exception ignored) {
+      return false;
+    }
```
orendain, line 63: The [only documented exception](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Boolean.html#getBoolean(java.lang.String)) this can throw is a SecurityException, which we should arguably not catch here, let alone catch and ignore.

Is there another reason this is consuming the exception?  If not, please consider simply returning `Boolean.getBoolean(...)`.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/Utils.java
```diff
@@ -42,4 +47,27 @@ public static String sanitizeBoolean(String value, DataField.DataTypeEnum dataTy
...
+  }
+
+  private static boolean isPredictionIntervalEnabledFromProperty() {
+    try {
+      return Boolean.parseBoolean(System.getenv(PREDICTION_INTERVAL_ENABLE_ENV_VAR));
+    } catch (Exception ignored) {
+      return false;
+    }
```
orendain, line 71: Similar to my last comment above, please avoid consuming arbitrary exceptions.  In this case, you can get by without any try-catch, propagating runtime exceptions (like `SecurityException`) if and when they come up.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/Utils.java
```diff
@@ -42,4 +47,27 @@ public static String sanitizeBoolean(String value, DataField.DataTypeEnum dataTy
     }
     return value;
   }
+
+  /**
+   * Check if prediction interval is enabled or not.
+   */
+  public static boolean isPredictionIntervalEnabled() {
+    return isPredictionIntervalEnabledFromEnv() || isPredictionIntervalEnabledFromProperty();
+  }
```
orendain, line 56: My two cents is the methods here are overkill. Consider placing all of the logic into a single function and keep it close to the place it needs to be used (not in `Utis.java`).

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -373,13 +374,31 @@ private static MojoPipeline loadMojoPipelineFromFile() {
...
     }
   }
+
+  private static MojoPipeline loadMojoPipelineFromFile(File mojoFile)
+      throws IOException, LicenseException {
+    try {
+      return MojoPipelineService.loadPipeline(mojoFile, buildPipelineConfig());
+    } catch (IllegalArgumentException e) {
+      log.debug("Disable Prediction interval because it is only supports regression model", e);
+      return MojoPipelineService.loadPipeline(mojoFile);
```
orendain, line 395: QQ: Is there a way to check support without expecting an exception (and one of generic type `IllegalArgumentException`?

If so, I'd recommend that check as a precondition, assuming there isn't any significant overhead.  Else, the catch-fallback LFTM is there is no other approach.

---

nit: the log message is a bit unclear.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -373,13 +374,31 @@ private static MojoPipeline loadMojoPipelineFromFile() {
...
+    } catch (Exception e) {
+      throw new RuntimeException("Unexpected exception occurs", e);
     }
   }
+
+  private static MojoPipeline loadMojoPipelineFromFile(File mojoFile)
+      throws IOException, LicenseException {
+    try {
+      return MojoPipelineService.loadPipeline(mojoFile, buildPipelineConfig());
+    } catch (IllegalArgumentException e) {
```
orendain, line 393: Instead of starting down this try-catch, **_and then_** checking if some env/property was set, it is recommended to instead explicitly first checking the env/property.

If not set, the flow would be simplified (no need for an additional try/catch)

If set, then you could be very explicit and fail without fallback if the MOJO2 runtime does not support the interval, rather than falling back.  (You can still fallback if that's a requirement, but doing so would no longer be coincidental).

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
```diff
@@ -373,13 +374,31 @@ private static MojoPipeline loadMojoPipelineFromFile() {
...
-      MojoPipeline mojoPipeline = MojoPipelineService.loadPipeline(mojoFile);
+      MojoPipeline mojoPipeline = loadMojoPipelineFromFile(mojoFile);
       log.info("Mojo pipeline successfully loaded ({}).", mojoPipeline.getUuid());
       return mojoPipeline;
     } catch (IOException e) {
       throw new RuntimeException("Unable to load mojo from " + mojoFile, e);
     } catch (LicenseException e) {
       throw new RuntimeException("License file not found", e);
+    } catch (Exception e) {
+      throw new RuntimeException("Unexpected exception occurs", e);
```
orendain, line 385: What's the reason for wrapping a generic `Exception` as a `RuntimeException` here?  Would the generic exception not otherwise still propagate up (and be assumed as an unexpected exception)?  Unless the wrapping is adding value (e.g., helpful message), I'd recommended against needing to do so.

orendain (changes requested):
Left a few notes.

Please be mindful that this repository is public, and links to private repositories like the one in the PR description may not be followed by all viewers.

Instead, link _to_ this PR _from_ those issues.  GitHub will still add contextual links in this PR for those that have access to the place those links were made in.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
jackjii79: @orendain the caveat is `MojoPipelineService` will explicit the exception if `prediction_interval` is unsupported for certain model and scorer does not able to do some pre-checking to find out if the model is supported or not

orendain: What the first part of my comment was getting at was that you could skip a lot of work in the case that the user does not want to prediction intervals.

E.g.:
```java
if ( noSpecialEnvVar ) {
  // Nothing special.
  return MojoPipelineService.loadPipeline();
}

// Only now must we do something new ...
```

In the first check, there is no `PipelineConfig` builder, no try/catch.  No introducing a different workflow than what already exists - a flow as straight as possible with minimal new places to look should something go wrong.

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
jackjii79: sounds fair

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/Utils.java
jackjii79: make sense

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/Utils.java
jackjii79: make sense

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/Utils.java
jackjii79: make sense

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
jackjii79: i could change to 'Disable Prediction interval as it is not supported for the given model' , wdyt ?

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
jackjii79: we could have three `loadPipeline()` in your flow:
```
if (noSpecialEnvVar) {
  return loadPipeline(file);
}

try {
  return loadPipeline(file, prediction_interval_enabled);
} catch (IllegalArgumentException e) {
  return loadPipelien(file);
}
```
But in my approach, only two is required, am i right ?

jackjii79:
@orendain PTAL, let me know if there is any red flag in your view so it wont get blocked

common/transform/src/main/java/ai/h2o/mojos/deploy/common/transform/MojoScorer.java
orendain: Correct, two paths in your example would be possible, but consider that this can be simplified down to one with some additional benefits below.

---

If there's no check for the envVar, then there's no guarantee (at least according to docs) that an `IllegalArgumentException` means simply that the user asked for prediction interval but the model does not support it.

In other words, the code will try load the pipeline, and if it fails for whatever reason due to catching an `IllegalArg...`, it will just try again without a config, regardless of if the config was the issue or not.

So ...

```java
try {
  return loadPipeline(file, prediction_interval_enabled);
} catch (IllegalArgumentException e) {
  return loadPipeline(file);
}
```

... will call `loadPipeline()` twice on _any_ `IllegalArgumentException`.  If the second `loadPipeline()` fails for any reason, then the second exception will propagate outward.

That means that if the issue was indeed that the envVar config was unsupported, that won't be the exception that makes it down the stack.  The only indicator of the config causing an issue is a debug message.

---

Consider if instead the code was:

```java
if (noSpecialEnvVar) {
  return loadPipeline(file);
}

return loadPipeline(file, prediction_interval_enabled);
```

or even simply:

```java
return loadPipeline(file, prediction_interval_enabled___OR___disabled);
```

... Then there would be a single path and the return exception would always be relevant.  The side effect either of the two above don't have is that a fallback is attempted.

Which prompts the question:  If the user has asked for a prediction range, and the scorer can't deliver it, are we sure the right move is to fallback to simply ignoring their configuration?  Or should we fail-fast?

Afterall, the new config is "PREDICTION_INTERVAL_ENABLED" and not something like "PREDICTION_INTERVAL_PREFERRED".

orendain (commented):
No red flags 👍🏾.  Commenting instead of approving as this is listed as a draft.

Just a bump to my comment about [removing links to private repositories](https://github.com/h2oai/dai-deployment-templates/pull/337#pullrequestreview-1400841603) as this repository is public.

jackjii79:
In favor of following PRs:
- https://github.com/h2oai/dai-deployment-templates/pull/351
- https://github.com/h2oai/dai-deployment-templates/pull/354
- https://github.com/h2oai/dai-deployment-templates/pull/355

PR #351, opened 2023-05-08
Issue 1756/extend prediction interval swagger schema

jackjii79:
https://github.com/h2oai/model-manager/issues/1756

Extend v1 schema with prediction interval support


jackjii79:
fyi @g-eoj updated as suggested here as well : https://github.com/h2oai/mlops-byom-images/pull/267#pullrequestreview-1421179673

common/swagger/v1/swagger.yaml
```diff
@@ -279,6 +288,10 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/Row'
+      requestPredictionIntervals:
+        type: boolean
+        description: |
+          If set to `true`. The scorer will try to fill field `predictionIntervals` in response if it is supported.
```
g-eoj, line 294: Should there be a comma? Like this
> If set to `true`, the scorer will try to fill field `predictionIntervals` in response if it is supported.

g-eoj (approved):
One comment. LGTM, thanks!

common/swagger/v1/swagger.yaml
jackjii79: thx, updated https://github.com/h2oai/dai-deployment-templates/pull/351/commits/f83117d0816b8434fc174e4cb0e5641375c47d4a

PR #355, opened 2023-05-11
Add new capability type to support prediction interval

jackjii79:
https://github.com/h2oai/model-manager/issues/1756
follow up of https://github.com/h2oai/dai-deployment-templates/pull/351

## Scope of changes
- Add SCORE_PREDICTION_INTERVAL as new capability type
- Port existing capability type from python runtime into swagger schema API


g-eoj (commented):
If we know that mojos don't support TTA, should we still add it to the swagger def in this repo?

jackjii79:
> If we know that mojos don't support TTA, should we still add it to the swagger def in this repo?

@g-eoj Technically all our runtime strictly follows the same swagger API, there is no reason not to make the swagger API cover all supported capability types, it is just happens to be in the same repo with java runtime does not mean it is only reference for this type of runtime, besides given this is a public repo i think it would also be useful to keep the API def up to date with the new features/update of runtime so as to make the swagger API as a complete reference to all our supported runtime and better instruct mlops user

PR #380, opened 2024-02-09
Change column not found log message to debug

g-eoj:
We constantly see this warning for legitimate scoring requests. 

It causes confusion when troubleshooting other issues.

I've never seen the warning help with actual issues.

Thus, we should not spam the logs with the warning as it causes more harm than good.


jackjii79 (approved):
lgtm

PR #385, opened 2024-03-12
Address memory issue caused by open api generator

jackjii79:
https://h2oai.slack.com/archives/G01C9KKQLAC/p1710175541471659

The root cause of performance issue is due to auto generated api model class should using `Row` class instead of `ArrayList<String>`, which makes JVM spend more than necessary heap size to do the class conversion results in slowness and abnormal memory increasing. 
This PR also removes deprecated sub gradle project from building pipeline


g-eoj (commented):
Thanks! I don't see any issues but prefer someone with more Java experience to give approval.

settings.gradle
```diff
@@ -14,18 +14,18 @@ dependencyResolutionManagement {
 }
 
 rootProject.name = 'dai-deployment-templates'
-include 'aws-lambda-scorer:lambda-template'
-include 'aws-lambda-scorer:terraform-recipe'
-include 'common:jdbc'
+//include 'aws-lambda-scorer:lambda-template'
```
jackjii79, line 17: All deprecated scorer which not only takes long time to build also no longer used by MLOps

jackjii79:
E2E: http://jenkins:8080/view/Model%20Ops/job/mlops/job/environments-helm/job/jh-helm/job/manual/job/test-mojo/allure/

shsma (approved):
LGTM! Thanks @jackjii79

PR #395, opened 2024-08-13
chore: Automate version specification

jakubhava:
- Remove publishing artifacts, we just need image output of this repo
- Update GCR -> GAR
- Automate versioning


.github/workflows/setup-environment.yml
```diff
@@ -49,9 +49,16 @@ jobs:
...
 
-      - name: Extract Gradle Component Version
+      - name: Save Component Version
         id: component_version
         run: |
-          componentVersion=$(grep version gradle.properties | grep -v '#' | sed -e 's/.*=//' | sed 's/^ *//')
-          echo "version: ${componentVersion}"
-          echo "version=${componentVersion}" >> $GITHUB_OUTPUT
+          if [[ "$GITHUB_REF" == refs/heads/main ]]; then
+            version="0.0.0-main.${{ steps.commit_hash.outputs.sha }}"
```
Rupeekshan, line 56: Why would we need this version since we only publishing images with `latest`, `latest-X.Y`, `vX.Y.Z` tags?

jakubhava: Internally, the component can print out the version as part of logs for example. It makes sense to print out the relevant output

Rupeekshan (approved):
LGTM!

PR #405, opened 2024-08-16
chore: Refine CICD with cleanups

Rupeekshan:


jakubhava (dismissed):

.github/workflows/image-security-scan.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - 'main'
       - 'release/[0-9]+.[0-9]+'
+    types: [ 'ready_for_review' ]
```
jakubhava, line 8: Why specifying the action type? I would simply run that scan as part of each run. Users can be fixing vulnerabilities with each commit

jakubhava: So this way we would optimise too much. We should fix here in all other repos

Rupeekshan: This filter was added recently based on running the security scan after the PR is in ready condition before merge. Otherwise many scan jobs will be triggered before PR is ready. For an example, scan jobs also on draft PR. Especially, in runtimes, that helped us to run scans efficiently since a lot of parallelism.

Rupeekshan: We can have this condition only on selective components which contains more parallel runs (i.e: runtimes, runtime-processors) and remove it from other repos. WDYT?

jakubhava: Let's keep it everywhere for now

jakubhava (approved):

PR #406, opened 2024-08-16
chore: Update component new name in references

Rupeekshan:


jakubhava (approved):
By changing the root project name some artifacts could be named differently

PR #421, opened 2024-09-08
chore: Switch to wolfi to fix critical vulnerability

jakubhava:


.github/workflows/setup-environment.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         run: |
           if ${{ github.event_name == 'pull_request' }}; then echo "sha=${{ github.event.pull_request.head.sha }}" >> $GITHUB_OUTPUT; fi
           if ${{ github.event_name == 'push' }}; then echo "sha=${{ github.sha }}" >> $GITHUB_OUTPUT; fi
+          if ${{ github.event_name == 'workflow_dispatch' }}; then echo "sha=$(git rev-parse --short=7 ${{ github.ref }})" >> $GITHUB_OUTPUT; fi
```
Rupeekshan, line 45: For my understanding, why do we shorten the character length to 7?

jakubhava: This is what I took from deployment server -> short sha is usually enough

Rupeekshan: Ok, in few repos, it looks like `git rev-parse --short ${{ github.ref }}`. 

To be consistent with all the repos, I'll update them to strict with 7 characters as part of https://github.com/h2oai/mlops/issues/1926.

senalw:
closes #423

